PR #5, opened 2021-11-08
NIAD-1854: MHS Mock

bbell-kainos:
Initial setup of MHS mock directories


dawidskainos (changes requested):
Could we rename it to "mhs-adaptor-mock"?

dawidskainos (changes requested):
I noticed one more thing - gp2gp translator and gpc-facade are on root level. Maybe let's stick to that?

ElliotB-96 (approved):

dawidskainos (approved):

mrogowska (approved):

PR #29, opened 2022-01-19
NIAD-1942: Create Codeable Concept mapper

mrogowska:


ElliotB-96 (approved):

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/CodeableConceptMapper.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    private boolean hasSnomedCodeInMainCode(CD codedData) {
+        return codedData.getCodeSystem().equals(SNOMED_SYSTEM_CODE);
+    }
+
+    private boolean hasSnomedCodeInTranslationElement(CD codedData) {
+        return codedData.getTranslation()
+            .stream()
+            .anyMatch(t -> t.getCodeSystem().equals(SNOMED_SYSTEM_CODE));
```
rorysproule1, line 30: More of a question here, but are we allowed to use single character variables like 't' in this case? For example would changing that to 'translation' be overkill?

rorysproule1 (approved):

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/CodeableConceptMapper.java
mrogowska: I've changed it to "translation", it will be more readable this way :)

ElliotB-96 (approved):

rorysproule1 (dismissed):

dawidskainos (approved):

rorysproule1 (approved):

PR #38, opened 2022-01-31
NIAD-1950: Generate ReferralRequest from RequestStatement

bbell-kainos:


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ReferralRequestMapper.java
```diff
@@ -0,0 +1,230 @@
...
+        var notes = getNotes(requestStatement);
+        var reasonCode = getReasonCode(requestStatement.getCode());
+        var authoredOn = getAuthoredOn(requestStatement.getAvailabilityTime());
+        var recipient = getRecipient(requestStatement.getResponsibleParty());
+        var requester = getRequesterAgentReference(requestStatement.getParticipant(), ehrComposition);
+
+        /**
+         * TODO: Implement future referencing
+         * - subject: references a global patient resource for the transaction
+         * - context: references an encounter resource if it has been generated from the ehrComposition
```
rorysproule1, line 55: Can you add the ticket number for each of these please? (sorry)

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ReferralRequestMapper.java
```diff
@@ -0,0 +1,230 @@
...
+         */
+
+        return createRequestStatement(id, identifier, notes, reasonCode, authoredOn, recipient, requester);
+    }
+
+    private Identifier getIdentifier(String id) {
+        Identifier identifier = new Identifier()
+                .setSystem(IDENTIFIER_SYSTEM) // TODO: concatenate source practice org id to URL
```
rorysproule1, line 63: and the ticker number for this (sorry again)

schema/src/main/java/org/hl7/v3/ObjectFactory.java
```diff
@@ -3509,4 +3510,9 @@ public RCMRIN030000UK06Subject createRCMRIN030000UK06Subject() {
     public JAXBElement<RCMRMT030101UK04EhrComposition> createEhrComposition(RCMRMT030101UK04EhrComposition value) {
         return new JAXBElement<RCMRMT030101UK04EhrComposition>(_EHR_COMPOSITION_QNAME, RCMRMT030101UK04EhrComposition.class, null, value);
     }
+
+    @XmlElementDecl(namespace = "", name = "RequestStatement")
+    public JAXBElement<RCMRMT030101UK04RequestStatement> createLocation(RCMRMT030101UK04RequestStatement value) {
```
rorysproule1, line 3515: I dont think you are using this are you? Your tests are all at the ehrComposition level? If you do need it then `createRequestStatement` instead of `createLocation`(if not then delete this and the QName on line 83 please

rorysproule1 (changes requested):

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ReferralRequestMapper.java
```diff
@@ -0,0 +1,230 @@
...
+            .map(II::getRoot)
+            .findFirst();
+    }
+
+    private boolean hasAgentReference(RCMRMT030101UK04Participant participant) {
+        return participant.getAgentRef() != null && participant.getAgentRef().getId() != null;
+    }
+
+    private boolean hasTypeCode(RCMRMT030101UK04Participant participant, String typeCode) {
+        return !participant.getTypeCode().isEmpty() && participant.getTypeCode().get(0).equals(typeCode);
```
mrogowska, line 146: Do we know if given typeCode is always a first element? Maybe it would be safer to check it this way:
```suggestion
        return participant.getTypeCode()
            .stream()
            .anyMatch(typeCode::equals);
```

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ReferralRequestMapper.java
```diff
@@ -0,0 +1,230 @@
...
+    private Reference getRecipient(RCMRMT030101UK04ResponsibleParty3 responsibleParty) {
+        if (responsiblePartyAgentRefHasIdValue(responsibleParty)) {
+            return new Reference(PRACTITIONER_REFERENCE_PREFIX
+                + responsibleParty.getAgentRef().getId().getRoot());
+        }
+        return null;
+    }
+
+    private boolean responsiblePartyAgentRefHasIdValue(RCMRMT030101UK04ResponsibleParty3 responsibleParty) {
+        return responsibleParty != null
```
mrogowska, line 91: I have the same concern as with checking typeCode inside hasTypeCode method, maybe it would be better to check it this way:
```suggestion
        return responsibleParty.getTypeCode().stream().anyMatch(RESP_PARTY_TYPE_CODE::equals)
            && responsibleParty.getAgentRef() != null
            && responsibleParty.getAgentRef().getId() != null;
```

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ReferralRequestMapper.java
```diff
@@ -0,0 +1,230 @@
...
+        Date authoredOn, Reference recipient, Reference requester) {
+        var referralRequest = new ReferralRequest();
+
+        referralRequest.setId(id);
+        referralRequest.getMeta().getProfile().add(new UriType(META_PROFILE));
+        referralRequest.getIdentifier().add(identifier);
+        referralRequest.setStatus(ReferralRequestStatus.UNKNOWN);
+        referralRequest.setIntent(ReferralCategory.ORDER);
+
+        if (requester != null) {
```
mrogowska, line 208: Could you check please (for example using unit tests), if setting null values for all of those fields can actually fail? If not, then maybe there's no need to check if they are null before setting them. Thanks!

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ReferralRequestMapper.java
```diff
@@ -0,0 +1,230 @@
...
+        if (availabilityTime != null) {
+            return DateFormatUtil.parse(availabilityTime.getValue()).getValue();
+        }
+        return null;
+    }
+
+    private Reference getRecipient(RCMRMT030101UK04ResponsibleParty3 responsibleParty) {
+        if (responsiblePartyAgentRefHasIdValue(responsibleParty)) {
+            return new Reference(PRACTITIONER_REFERENCE_PREFIX
+                + responsibleParty.getAgentRef().getId().getRoot());
```
mrogowska, line 85: I think this line can be moved to the line above it, it won't be too long :)

mrogowska (changes requested):

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ReferralRequestMapper.java
bbell-kainos: Unfortunately I need the `responsibleParty != null` check but i've added in your suggestion

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ReferralRequestMapper.java
bbell-kainos: You're right, these null checks aren't needed! 😄

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ReferralRequestMapper.java
```diff
@@ -0,0 +1,216 @@
...
+    private Identifier getIdentifier(String id) {
+        Identifier identifier = new Identifier()
+            .setSystem(IDENTIFIER_SYSTEM) // TODO: concatenate source practice org id to URL (NIAD-2021)
+            .setValue(id);
+        return identifier;
+    }
+
+    private CodeableConcept getReasonCode(CD reasonCode) {
+        if (reasonCode != null) {
+            return new CodeableConceptMapper().mapToCodeableConcept(reasonCode);
```
mrogowska, line 71: I'm sorry Barry, I've asked Elliot for the same thing in his PR, but I'm not sure which one will be merged first, so I'll add this here as well :D If you could annotate CodeableConceptMapper class with @Service annotation, and inject it in your mapper, instead of doing`new CodeableConceptMapper()`? I forgot to add this annotation to CodeableConceptMapper earlier :( And then you could mock it in ReferralRequestMapperTest

mrogowska (changes requested):

mrogowska (approved):

PR #55, opened 2022-02-08
NIAD-1946: Generate immunisation from observation statement

AlanSimpson1723:


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,171 @@
...
+        return null;
+    }
+
+    private Identifier getIdentifier(String id) {
+        return new Identifier()
+            .setSystem(IDENTIFIER_SYSTEM) // TODO: concatenate source practice org id to URL (NIAD-2021)
+            .setValue(id);
+    }
+
+    private boolean checkSnomedCode(RCMRMT030101UK04ObservationStatement observationStatement) {
```
dawidskainos, line 79: I think it's a good idea to prefix methods that return booleans with 'is'. This will automatically force you to name it in a more specific way.
'check' is fine for checking conditions and throwing exceptions if mot matched.

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,171 @@
...
+        String snomedCode = observationStatement.getCode().getCodeSystem();
+
+        return SNOMED_CODE.equals(snomedCode);
+    }
+
+    private Extension createExtension(RCMRMT030101UK04ObservationStatement observationStatement) {
```
dawidskainos, line 85: more specific method name maybe? what kind of extension?

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,171 @@
...
+        return new Extension()
+                    .setUrl(VACCINE_PROCEDURE_URL)
+                    .setValue(new Extension()
+                        .setUrl(URL)
+                        .setValue(codeableConceptMapper.mapToCodeableConcept(observationStatement.getCode()))
+                    );
+    }
+
+    private Date getDate(RCMRMT030101UK04ObservationStatement observationStatement) {
```
dawidskainos, line 94: What date is that? Method name could be more specific

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    private List<Immunization> createImmunization(String id, Identifier identifier, List<Annotation> note,
+        Extension vaccineExtension, Date date, Extension recordedTimeExtension) {
+        var immunization = new Immunization();
+
+        immunization.getMeta().getProfile().add(new UriType(META_PROFILE));
+        immunization.getIdentifier().add(identifier);
+
+        try {
```
dawidskainos, line 153: We should't use kind of exception driven approach. I'm not really sure what are you trying to achieve here - definitely needs rework.

dawidskainos (changes requested):
I haven't reviewed tests yet as there is still some work to be done first.

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,171 @@
...
+        } else if (observationStatement.getEffectiveTime().getCenter() != null) {
+            return DateFormatUtil.parse(observationStatement.getEffectiveTime().getCenter().getValue()).getValue();
+        } else if (observationStatement.getEffectiveTime().getLow() != null) {
+            return DateFormatUtil.parse(observationStatement.getEffectiveTime().getLow().getValue()).getValue();
+        } else if (observationStatement.getEffectiveTime().getHigh() != null
+            && observationStatement.getAvailabilityTime() != null
+            && observationStatement.getAvailabilityTime().getValue() != null) {
+            return DateFormatUtil.parse(observationStatement.getAvailabilityTime().getValue()).getValue();
+        } else if (observationStatement.getEffectiveTime() == null && observationStatement.getAvailabilityTime().getValue() != null) {
+            return DateFormatUtil.parse(observationStatement.getAvailabilityTime().getValue()).getValue();
```
rorysproule1, line 101: Do me a favour and check each of these values with nullFlavor="UNK" insetad of a value. I suspect this might throw an error. My current PR has some functionality which checks for this if you need a reference https://github.com/NHSDigital/nia-patient-switching-standard-adaptor/pull/57/files#diff-fadb5d3a335b0ecd8a6396f0e489d8ff26a4ad6d8f517e561c710d17f9da8600R47

rorysproule1 (changes requested):

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,171 @@
...
+    }
+
+    private Identifier getIdentifier(String id) {
+        return new Identifier()
+            .setSystem(IDENTIFIER_SYSTEM) // TODO: concatenate source practice org id to URL (NIAD-2021)
+            .setValue(id);
+    }
+
+    private boolean hasImmunizationCode(RCMRMT030101UK04ObservationStatement observationStatement) {
+        String snomedCode = observationStatement.getCode().getCodeSystem();
```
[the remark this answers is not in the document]
AlanSimpson1723: cardinality is 1.1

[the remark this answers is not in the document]
AlanSimpson1723: > Does ObservationStatement always have a code && codeSystem? If it doesn't then add a check for these

cardinality is 1.1

gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapperTest.java
```diff
@@ -0,0 +1,164 @@
...
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import lombok.SneakyThrows;
+
+@ExtendWith(MockitoExtension.class)
+public class ImmunizationMapperTest {
```
rorysproule1, line 27: Could you also add a test that tests when every optional field in the ObservationStatement isn't provided?

rorysproule1 (changes requested):

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,301 @@
...
+        var practitioner = ParticipantReferenceUtil.getParticipantReference(observationStatement.getParticipant(), ehrComposition);
+        var encounter = getEncounterReference(encounterList, ehrComposition.getId());
+
+        immunizationMapperParameters.setIdParam(id);
+        immunizationMapperParameters.setIdentifierParam(identifier);
+        immunizationMapperParameters.setNoteParam(note);
+        immunizationMapperParameters.setDateParam(date);
+        immunizationMapperParameters.setRecordedTimeExtensionParam(recordedTimeExtension);
+        immunizationMapperParameters.setVaccineExtensionParam(vaccineExtension);
+        immunizationMapperParameters.setPatientParam(new Reference(patientResource.getIdElement()));
```
rorysproule1, line 91: Sorry Alan, probably explained badly from me, all you need to do here is change this parameter to accept a Patient instead of a reference and pass the patientResource. That should eeal with everything itself (so no need to create any Reference object)

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,301 @@
...
+                return new Reference(matchingEncounter.get().getIdElement());
+            }
+        }
+
+        return null;
+    }
+
+    private boolean matchingIds(String encounterId, II ehrCompositionId) {
+        return encounterId.equals(ehrCompositionId.getRoot());
+    }
```
rorysproule1, line 114: Just a favour to ask with this one, get in contact with Bart C on this, i know he is moving it to a Util class so depending on how merges first, if you do then just ask him to change this to instead use his UtilClass please!

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,301 @@
...
+        private Extension vaccineExtensionParam;
+        private Date dateParam;
+        private Extension recordedTimeExtensionParam;
+        private Reference patientParam;
+        private Reference encounterParam;
+        private Reference practitionerParam;
+
+        public ImmunizationMapperParameters() {
+
+        }
```
rorysproule1, line 299: Is this needed?

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/service/BundleMapperService.java
```diff
@@ -61,10 +65,17 @@ public Bundle mapToBundle(RCMRIN030000UK06Message xmlMessage) {
         var referralRequests = mapReferralRequests(ehrFolder, patient);
         addEntries(bundle, referralRequests);
 
+        var immunizations = mapImmunizations(ehrExtract, patient, List.of(new Encounter()));
```
rorysproule1, line 68: Another favour to ask, can you leave a comment here

// TODO: Insert encounter list (NIAD-1961)

For myself to link up later!

rorysproule1 (changes requested):

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
AlanSimpson1723: > Is this needed?

yea i believe so as this is the constructor that actually allows me to call this class but ill check if there is another way of doing this

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,301 @@
...
+                return DateFormatUtil.parseToDateTimeType(low).getValue();
+            } else if (high != null && availabilityTimeValue.getValue() != null && availabilityTimeValue.getNullFlavor() == null) {
+                return DateFormatUtil.parseToDateTimeType(availabilityTimeValue.getValue()).getValue();
+            }
+        }
+
+        return null;
+    }
+
+    private String highValueToNotes(RCMRMT030101UK04ObservationStatement observationStatement) {
```
boydstep, line 205: I think the name of this function is confusing - from the name - I would expect it to add to the notes. Since we're just returning a string value perhaps we should change the name, maybe something like `createDateText`?

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,301 @@
...
+        var dateText = highValueToNotes(observationStatement);
+
+        if (pertinentText != null && dateText != null) {
+            var noteStringType = new StringType(pertinentText + StringUtils.SPACE + dateText);
+            return new Annotation(noteStringType);
+        } else if (pertinentText != null) {
+            var noteStringType = new StringType(pertinentText);
+            return new Annotation(noteStringType);
+        } else if (dateText != null) {
+            var noteStringType = new StringType(dateText);
```
boydstep, line 234: Minor - but rather than creating the extra variables - we could just do: `return new Annotation(new StringType(dateText))`

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,301 @@
...
+            }
+        }
+
+        return null;
+    }
+
+    private Immunization createImmunization(ImmunizationMapperParameters immunizationMapperParameters) {
+        var immunization = new Immunization();
+
+        immunization.getMeta().getProfile().add(new UriType(META_PROFILE));
```
boydstep, line 256: Since we're working on a new Immunization object and won't have any data in the meta field, we can just set the profile here instead of getting & then adding to it.

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
AlanSimpson1723: > Just a favour to ask with this one, get in contact with Bart C on this, i know he is moving it to a Util class so depending on how merges first, if you do then just ask him to change this to instead use his UtilClass please!

Bart, got back to me and said he no longer needs this in his class has there is a bit of work around it so i am good to keep this in, thanks

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,257 @@
...
+        immunization.getExtension().add(createVaccineProcedureExtension(observationStatement));
+        immunization.getExtension().add(createRecordedTimeExtension(ehrComposition));
+        immunization
+            .setStatus(Immunization.ImmunizationStatus.COMPLETED)
+            .setNotGiven(false)
+            .setPrimarySource(false)
+            .setPatient(new Reference(patientResource))
+            .setId(id);
+
+        if (date != null) {
```
dawidskainos, line 90: I think all these null checks are not necessary. HAPI is nullsafe.

gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapperTest.java
```diff
@@ -0,0 +1,197 @@
...
+    @Test
+    public void mapObservationToImmunizationWithHighEffectiveTime() {
+        var ehrExtract = unmarshallEhrExtract("immunization_with_only_high_effective_time.xml");
+        List immunizationList = immunizationMapper.mapToImmunization(ehrExtract, getPatient(), getEncounterList());
+
+        var immunization = (Immunization) immunizationList.get(0);
+        assertImmunizationWithHighEffectiveTime(immunization);
+    }
+
+    private void assertImmunizationWithHighAndLowEffectiveTime(Immunization immunization) {
```
dawidskainos, line 116: Some tests are timezone dependant - they're failing if timezone is different than GMT.

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,257 @@
...
+        } else if (ivxbts.getValue() != null) {
+            return ivxbts.getValue();
+        } else if (ivxbts.getNullFlavor().equals(CsNullFlavor.UNK)) {
+            return CsNullFlavor.UNK.value();
+        }
+
+        return null;
+    }
+
+    private Date getObservationDate(RCMRMT030101UK04ObservationStatement observationStatement) {
```
dawidskainos, line 192: please don't use java.util.Date - it looses timezone information and will cause test failures

dawidskainos (changes requested):

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,248 @@
...
+    private String getTSStringValue(TS ts) {
+        if (ts == null) {
+            return null;
+        } else if (ts.getValue() != null) {
+            return ts.getValue();
+        } else if (ts.getNullFlavor().equals(CsNullFlavor.UNK)) {
+            return CsNullFlavor.UNK.value();
+        }
+
+        return null;
```
mrogowska, line 168: You could simplify it by checking if ts is not null first, then you'll only have to return null once:
```suggestion
        if (ts != null) {
            if (ts.getValue() != null) {
                return ts.getValue();
            } else if (ts.getNullFlavor().equals(CsNullFlavor.UNK)) {
                return CsNullFlavor.UNK.value();
            }
        }

        return null;
```

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,248 @@
...
+    }
+
+    private String getIVXBTSStringValue(IVXBTS ivxbts) {
```
mrogowska, line 171: You could change this method according to the suggestion from `getTSStringValue(TS ts)` as well

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,248 @@
...
+            return ivxbts.getValue();
+        } else if (ivxbts.getNullFlavor().equals(CsNullFlavor.UNK)) {
+            return CsNullFlavor.UNK.value();
+        }
+
+        return null;
+    }
+
+    private DateTimeType getObservationDate(RCMRMT030101UK04ObservationStatement observationStatement) {
+        if (observationStatement.getEffectiveTime() != null) {
```
mrogowska, line 184: I would create a variable for `observationStatement.getEffectiveTime()` as it's repeated a few times later

gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapperTest.java
```diff
@@ -0,0 +1,202 @@
...
+
+    @BeforeEach
+    public void setup() {
+        setUpCodeableConceptMock();
+    }
+
+    @Test
+    public void mapObservationToImmunizationWithValidData() {
+        var ehrExtract = unmarshallEhrExtract("full_valid_immunization.xml");
+        List immunizationList = immunizationMapper.mapToImmunization(ehrExtract, getPatient(), getEncounterList());
```
mrogowska, line 54: Please change the type of immunizationList (in all tests) to `List<Immunization>`

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/util/EhrResourceExtractorUtil.java
```diff
@@ -37,6 +38,30 @@ private static boolean validPlanStatement(RCMRMT030101UK04Component4 component,
         return component.getPlanStatement() != null && component.getPlanStatement().getId() == resourceId;
     }
 
+    public static List<RCMRMT030101UK04EhrComposition> extractValidImmunizationEhrCompositions(RCMRMT030101UK04EhrExtract ehrExtract) {
```
mrogowska, line 41: You could move this public method higher, so all the public methods are above the private methods.

mrogowska (changes requested):

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ImmunizationMapper.java
```diff
@@ -0,0 +1,248 @@
...
+        immunization
+            .setDateElement(date)
+            .setEncounter(encounter)
+            .addPractitioner(new Immunization.ImmunizationPractitionerComponent(practitioner))
+            .setStatus(Immunization.ImmunizationStatus.COMPLETED)
+            .setNotGiven(false)
+            .setPrimarySource(false)
+            .setPatient(new Reference(patientResource))
+            .setId(id);
+        immunization.getNote().add(note);
```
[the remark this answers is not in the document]
dawidskainos: I think you could just something like `immunization.addNote(buildNote(observationStatement));`
HAPI is very null safe

PR #77, opened 2022-03-01
NIAD-2046 - Quick fix of null handling

bchotkowski:
-> Secured an optional field from passing null


bchotkowski:
Moved to one-commit-pr -> #78

PR #91, opened 2022-03-03
NIAD-2021: Concatenate Organisation ID to identifier.system

bbell-kainos:


rorysproule2 (approved):

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ObservationCommentMapper.java
```diff
@@ -96,21 +94,6 @@ private void setObservationComment(Observation observation, String text) {
         }
     }
 
-    private Meta createMeta() {
```
mrogowska, line 99: Is this no longer needed?

mrogowska (changes requested):

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ObservationCommentMapper.java
bbell-kainos: No they are not, meta and identifier fields are generated in the ResourceUtil class as it is repeated functionality across many classes

mrogowska: I see, thanks!

mrogowska (approved):

rorysproule2 (approved):

PR #99, opened 2022-03-08
NIAD-1944: Implement Codeable Concept Mapper using Snomed CT Database

bbell-kainos:


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/CodeableConceptMapper.java
```diff
@@ -40,22 +177,80 @@ private CD getSnomedTranslationElement(CD codedData) {
...
             .getCodingFirstRep()
             .setCode(code)
             .setSystem(system)
-            .setDisplay(display);
+            .setDisplay(display)
+            .addExtension(extension);
         return codeableConcept;
     }
+
+    private SnomedCTDescription getSnomedDescriptionUsingConceptIdDisplayName(String conceptId, String displayName) {
```
dawidskainos, line 213: I think that private methods that do nothing more than invoke another method on snomedCTDao are not necessary

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/CodeableConceptMapper.java
```diff
@@ -1,25 +1,162 @@
...
+    private static final String DESCRIPTION_PARTITION_SHORT = "01";
+    private static final String DESCRIPTION_PARTITION_LONG = "11";
+    private static final int MAX_CHARACTERS_FROM_RIGHT = 3;
+    private static final int MIN_CHARACTERS_FROM_RIGHT = 1;
 
-    public CodeableConcept mapToCodeableConcept(CD codedData) {
+    @Autowired
+    private final SnomedCTDao snomedCTDao;
+
+    public CodeableConcept mapToCodeableConcept(CD codedData, boolean isMedicationResource) {
```
mrogowska, line 36: I think it would be better to create a separate method for medication resource. There's only one case when isMedicationResource is true and you need to pass false each time - better to have 2 public methods.

docker/vars.local.tests.sh
```diff
@@ -3,6 +3,9 @@
 export PS_DB_URL="jdbc:postgresql://ps_db:5432"
 export PS_DB_OWNER_NAME="postgres"
 export PS_DB_OWNER_PASSWORD="pass_test"
+export PS_DB_HOST="ps_db"
```
mrogowska, line 6: If we need to have separate variables for db's host and port, then we could remove PS_DB_URL variable and create the url from host and port where needed (in db-connector/build.gradle and application.yml files)

mrogowska (changes requested):

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/CodeableConceptMapper.java
```diff
@@ -1,25 +1,162 @@
...
             var translation = getSnomedTranslationElement(codedData);
-            var display = getDisplayForSnomedTranslationElement(codedData, translation);
-            return createCodeableConcept(translation.getCode(), SNOMED_SYSTEM, display, codedData.getOriginalText());
+            return generateSnomedCodeableConcept(translation, codedData, isMedicationResource);
+        }
+
+        return new CodeableConcept().setText(generateText(codedData.getOriginalText(), codedData.getDisplayName()));
+    }
+
+    private CodeableConcept generateSnomedCodeableConcept(CD codedData, CD mainCodeFromTranslation, boolean isMedicationResource) {
```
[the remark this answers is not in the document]
bbell-kainos: Tried a few approaches there to cut the method down although ran into an issue with each. Figured the best approach was to make the function more readable- but happy to discuss :)

docker/vars.local.tests.sh
```diff
@@ -1,8 +1,10 @@
 #!/usr/bin/env bash
-
-export PS_DB_URL="jdbc:postgresql://ps_db:5432"
+export PS_DB_HOST="ps_db"
+export PS_DB_PORT="5432"
+export PS_DB_URL="jdbc:postgresql://${PS_DB_HOST}:${PS_DB_PORT}"
 export PS_DB_OWNER_NAME="postgres"
 export PS_DB_OWNER_PASSWORD="pass_test"
+export PGPASSWORD=`echo $PS_DB_OWNER_PASSWORD`
 export POSTGRES_PASSWORD=`echo $PS_DB_OWNER_PASSWORD`
```
dawidskainos, line 8: Do we really need 3 vars with the same value?

mrogowska (dismissed):

mrogowska (approved):

dawidskainos (approved):

PR #101, opened 2022-03-08
NIAD-2023: Template Mapper

rorysproule2:


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/service/BundleMapperService.java
```diff
@@ -125,6 +130,11 @@ public Bundle mapToBundle(RCMRIN030000UK06Message xmlMessage) {
         return bundle;
     }
 
+    private List<? extends DomainResource> mapTemplates(RCMRMT030101UK04EhrExtract ehrExtract, Patient patient,
+        List<Encounter> encounters, String practiseCode) {
+        return templateMapper.mapTemplates(ehrExtract, patient, encounters, practiseCode);
```
mrogowska, line 135: I don't think this method is needed, you can just call `templateMapper.mapTemplates(ehrExtract, patient, encounters, practiseCode);` inside mapToBundle method. (Elliot is doing the cleanup like this in his PR so it'll be nice to keep it consistent from now on)

mrogowska (changes requested):
It might be good if you and Elliot decide whose PR will be merged first, as Elliot's done some cleanup in his, and there will be some sad conflicts :(

mrogowska (dismissed):

mrogowska (approved):

bchotkowski (approved):

PR #112, opened 2022-03-25
NIAD-2009: Implement TLS MA

mrogowska:


docker/example.vars.sh
```diff
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 
-export PS_DB_URL="jdbc:postgresql://localhost:5436"
 export PS_DB_OWNER_NAME="postgres"
```
bbell-kainos, line 3: Do we need PS_DB_URL to look like the following? (Apologies this is something I possibly could have changed in my PR!)

```suggestion
export PS_DB_OWNER_NAME="postgres"
export PS_DB_HOST="localhost"
export PS_DB_PORT="5436"
export PS_DB_URL="jdbc:postgresql://${PS_DB_HOST}:${PS_DB_PORT}"
```

bbell-kainos (changes requested):

docker/example.vars.sh
mrogowska: I've removed this variable, because it wasn't used, as it's overwritten in the docker-compose.yml anyway:
PS_DB_URL=jdbc:postgresql://ps_db:5432 (my bad, I should have removed it earlier). PS_DB_HOST and PS_DB_PORT are not required here, as they are used in the snomed scripts only, but if you think it would be useful to have them here, then I can add them

bbell-kainos: Ah thats not a problem, approved!

bbell-kainos (approved):

bchotkowski (approved):

PR #128, opened 2022-04-12
Windows environment

SRAlexander:


docker/example.vars.sh
```diff
@@ -1,15 +1,33 @@
...
+export PS_DB_PORT='5436';
+export HOSTNAME='localhost';
+
+export PS_DB_URL="jdbc:postgresql://$HOSTNAME:$DB_PORT";
+export PS_DB_OWNER_NAME="postgres";
+export PS_FROM_ODS_CODE="PSS_001";
+export PS_DB_OWNER_PASSWORD="123456";
+export PS_DB_HOST="$HOSTNAME";
+export POSTGRES_PASSWORD="$PS_DB_OWNER_PASSWORD";
+export GPC_FACADE_USER_DB_PASSWORD="123456";
```
SRAlexander, line 13: More passwords that should be empty by default

PR #152, opened 2022-05-13
Niad 2036 manage extract timeout

benhession:
## Requirement

As a consumer of PSS adaptor
I want the adaptor to handle EHR Response Timeout
So that the electronic EHR Extract process could be failed and paper copy fallback applied by losing practice

## Acceptance Criteria

- PersistDuration must be taken from SDS API via /Endpoint endpoint

- nhsMHSPersistDuration is the url parameter in FHIR response

- Register PS Adaptor as SDS client

- Parse ISO 8601 duration format

- Calculate timeout as per spec:Timeout [secs]= (A x persistDuration contract property of EHR Response [secs]) + (B x Number of COPC Common Point to Point EHR messages x persistDuration contract property of COPC Common Point to Point messages [secs])

- “Cron-like” job to detect timed out transfers and send NACK (response code 25)

- A & B - Will be determined by the end user so should be configurable parameters (Scale not seconds).
- Count of COPC messages must be dynamic, original count from RM06 message + 1 for every MID reference in further COPC messages. 
- Cron job - running schedule to be determined by timescales.

- polling SDS will not occur on every poll but every x polls to deal with the potential of it changing.


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/task/scheduled/EHRTimeoutHandler.java
```diff
@@ -0,0 +1,133 @@
...
+            LOGGER.error("Error retrieving persist duration: [{}]", e.getMessage());
+        } catch (JsonProcessingException | SAXException | DateTimeParseException | JAXBException e) {
+            LOGGER.error("Error parsing inbound message from database");
+            migrationStatusLogService.addMigrationStatusLog(EHR_GENERAL_PROCESSING_ERROR, conversationId);
+        }
+
+        mdcService.applyConversationId("");
+    }
+
+    private void sendNackMessage(InboundMessage inboundMessage, String conversationId) throws JAXBException {
```
ellenwootten, line 114: is there a util for this? i feel like i've seen this lots of times now!

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/util/OutboundMessageUtil.java
```diff
@@ -0,0 +1,39 @@
+package uk.nhs.adaptors.pss.translator.util;
+
+import org.hl7.v3.RCMRIN030000UK06Message;
+import org.springframework.stereotype.Component;
+
+@Component
+public class OutboundMessageUtil {
```
ellenwootten, line 7: i think this is all in xml parse util? could be the other type of payload though. I think this makes more sense as a class and title to me though.

ellenwootten (approved):
looks really good to me - had a couple of questions but i am happy 👍

SRAlexander (approved):

PR #156, opened 2022-05-25
NIAD-2154: insert narrative statement into skeleton - BUG FIX

benhession:
### Description

When a skeleton message is replaced by the complete EHR Extract the additional Narrative Statement is not inserted into the skeleton correctly.

### DIFF
--- a/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/service/InboundMessageMergingService.java
+++ b/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/service/InboundMessageMergingService.java
@@ -7,6 +7,7 @@ import static uk.nhs.adaptors.pss.translator.util.XmlUnmarshallUtil.unmarshallSt
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Optional;
 
 import javax.xml.bind.JAXBException;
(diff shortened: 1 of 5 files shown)


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/service/InboundMessageMergingService.java
```diff
@@ -69,19 +68,18 @@ private void findAndReplaceSkeleton(List<PatientAttachmentLog> attachmentLogs, I
             throws SAXException, TransformerException {
         // merge skeleton message into original payload
         var skeletonLogs = attachmentLogs.stream().filter(log -> log.getSkeleton().equals(true)).toList();
-        var skeletonFileName = skeletonLogs.stream().findFirst().get().getFilename();
+        var skeletonFileName = skeletonLogs.stream().findFirst().map(PatientAttachmentLog::getFilename).orElseThrow();
```
SRAlexander, line 71: Do we need to update any unit tests with these modifications?

benhession: The `map` + `orElseThrow` was introduced to avoid the unchecked `get()`. I've refactored the code so throwing the exception is no longer required.

SRAlexander (approved):

PR #157, opened 2022-05-25
NIAD-2155: Update default CRON schedule - BUG FIX

benhession:
### Description

The current CRON schedule should be every six hours, but is currently every second of the hour, every six hours.

### DIFF
--- a/gp2gp-translator/src/main/resources/application.yml
+++ b/gp2gp-translator/src/main/resources/application.yml
@@ -51,7 +51,7 @@ sds:
 timeout:
   ehrExtractWeighting: ${TIMEOUT_EHR_EXTRACT_WEIGHTING:1}
   copcWeighting: ${TIMEOUT_COPC_WEIGHTING:1}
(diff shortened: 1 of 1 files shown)


gp2gp-translator/src/main/resources/application.yml
```diff
@@ -51,7 +51,7 @@ sds:
 timeout:
   ehrExtractWeighting: ${TIMEOUT_EHR_EXTRACT_WEIGHTING:1}
   copcWeighting: ${TIMEOUT_COPC_WEIGHTING:1}
-  cronTime: ${TIMEOUT_CRON_TIME:* * */6 * * *}
+  cronTime: ${TIMEOUT_CRON_TIME:1 1 */6 * * *}
```
SRAlexander, line 54: Why not just 0's?

benhession: Updated

SRAlexander (approved):

PR #171, opened 2022-06-22
task_2161

SRAlexander:
… refactoring of code


adonaimartins (commented):
filename = filename.replace("%20", " ");
and then 
filename = filename.replace(" ", "%20");

is innecesary replacing the spaces again " " with "%20". it would be easier creating a new var with:

tempFileWithCodeSpaces =  filename.replace("%20", " ");

maybe??

gp2gp-translator/src/integrationTest/java/uk/nhs/adaptors/pss/translator/E2EMappingIT.java
```diff
@@ -207,7 +211,7 @@ private String parseMessageToString(InboundMessage inboundMessage) {
 
     private boolean isEhrExtractTranslated() {
```
adonaimartins, line 212: this method, isEhrExtractTranslated() is repeated and the other method that does the same is called "isEhrMigrationCompleted". one of the methods should be picked and the other method should be deleted

SRAlexander: Filename issue is reversed so we know the process is correct.

PR #172, opened 2022-07-01
Task 2195

SRAlexander:


docker/release-scripts/ps-mhs-e2e/docker-compose.yml
```diff
@@ -0,0 +1,178 @@
...
+  ############### MHS specific   ###################
+
+  inbound:
+    image: nhsdev/nia-mhs-inbound:1.2.2
+    networks:
+      - nia-ps
+    ports:
+      - "443"
+      - "80"
+      - "${MHS_INBOUND_HEALTHCHECK_SERVER_PORT}:${MHS_INBOUND_HEALTHCHECK_SERVER_PORT}"
```
adonaimartins, line 20: the port on the left of the colon ":" is the port the NHS will hit when they try to access it. the port on the right is the port inside the docker environment. within the same network, they should not be the same and I would hardcode them so the user outside of the network, sees this port as available and choses it, and it happens to be used inside docker. 

the inner port could be hardcoded? they could pick one of the ports used in here, example (8085) and it would be in use. I would hardcoded it like:
   ports:
      - "443"
      - "80"
      - "${MHS_INBOUND_HEALTHCHECK_SERVER_PORT}:8083"
      
      I would also check that any ports after the colon ":" don't match with any of the other ports in the service

PR #173, opened 2022-07-01
NIAD-2177 bug fixed and integration test fixed

SRAlexander:


snomed-immunization-loader/load_immunization_codes.sh
```diff
@@ -30,4 +30,4 @@ then
 fi
 
 databaseUri="postgresql://${PS_DB_OWNER_NAME}:${POSTGRES_PASSWORD}@${PS_DB_HOST}:${PS_DB_PORT}/${dbName}"
-psql ${databaseUri} -c "\copy ${snomedCtSchema}.immunization_codes (conceptid, description, safetycode) FROM '${basedir}/$1' DELIMITER ',' CSV HEADER QUOTE '\"'"
\ No newline at end of file
+psql ${databaseUri} -c "\copy ${snomedCtSchema}.immunization_codes (conceptid, description, safetycode) FROM '${basedir}/snomed_immunization_codes.csv' DELIMITER ',' CSV HEADER QUOTE '\"'"
```
SRAlexander, line 33: This needs reverting back to the original version. It seems to work but I don't want to hard code this area incase we swap variables out in Jenkins later down the line.

PR #176, opened 2022-07-12
task_2191 quantity value setter updated to handle decimals and updated expect…

SRAlexander:
…ed test data where required


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/QuantityMapper.java
```diff
@@ -69,6 +70,6 @@ private void setQuantityWithLowComparator(Quantity quantity, PQInc low) {
 
     private void setQuantityValueAndUnit(Quantity quantity, String value, String unit, List<PQR> translation) {
         setUnit(quantity, unit, translation);
-        quantity.setValue((long) Double.parseDouble(value));
+        quantity.setValue(new BigDecimal(Double.parseDouble(value)));
```
ellenwootten, line 73: is there any reason why setValue can't use the double? / or we can't use either new BigDecimal(value) or BigDecimal.valueOf(value) ?

ellenwootten: confirmed, nothing simpler passes the tests.
`Double.parseDouble(value)` fails as so: `expected: 27 but was: 27.0`
`new BigDecimal(value)` fails as so: `expected: 27 but was: 27.000`
Strange! Approved review :)

ellenwootten (approved):

PR #190, opened 2022-08-08
Document reference IDs now in expected output (2200)

rio-k9:


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ConditionMapper.java
```diff
@@ -260,14 +264,32 @@ private CodeableConcept generateCategory() {
...
-            .flatMap(statementRef -> bundle.getEntry()
-                .stream()
-                .filter(entryComponent -> statementRef.getId().getRoot().equals(entryComponent.getResource().getId()))
-            )
-            .map(BundleEntryComponent::getResource)
-            .map(resource -> buildReferenceExtension(RELATED_CLINICAL_CONTENT_URL, new Reference(resource)))
+
+        // Filter for bundle entries where entry ID exists in both streams
+        var bundleIdArr = bundle.getEntry()
+            .stream()
```
roberttuppenanswerdigital, line 270: Possibly move each of these into a method so it explains what each are trying to do?

roberttuppenanswerdigital (changes requested):

PR #208, opened 2022-09-23
NIAD-2336 testGroup 6 added

adonaimartins:


SRAlexander (changes requested):
Test 02 - B: Verify continue message. First test fails The message body should contain fragment MID. | AssertionError: expected false to be true.

There are references to a function called retryRequest that have no function declared in the test script which causes errors if the check case fails e.g... 01 - B Verify EHR Request. PLease check all similar functions that the retryRequest function exists.

SRAlexander (approved):

PR #216, opened 2022-12-01
NIAD-2466: Bug-Fix façade returns 500 after COPC received

benhession:
- Added additional migration statuses to the facade. 

- Renamed COPC message processing statuses as it is easy to confuse "continue message" with the continue message sent to the incumbent.   

- Added COPC_ACKNOWLEDGED migration status, as the EHR_REQUEST_ACKNOWLEDGED status was being used for COPC messages.


benhession:
In the case of a negative ack from the incumbent, the façade responded with the unsupported method response, consisting of a status code of 500 and a body containing a diagnostic message "please contact an administrator". As the 500 response is expected in this scenario, I have added support for the migration status of EHR_REQUEST_NEGATIVE_ACK, which now returns a 500 and an empty body.

PR #228, opened 2023-01-25
NIAD-2359 additional list entries

benhession:
**Investigation**
In the case of a structured EHR Composition (with Topics and Categories), mapped Conditions were refenced as entries to newly created Topics in the ASR. These were identified as the additional list entries. 

**Work Undertaken**

The Conditions are now mapped to the appropriate Topics in the ASR. The appropriate Topic is determined by the location of the nested statement referenced in the `conditionNamed` field of the Link Set the condition is mapped from.


SRAlexander (approved):
Looks good from here :)

PR #230, opened 2023-01-26
Niad-2509: bugfix - medication request prepended true

benhession:
**Description**

We were expecting the term Expired to be here without the “True“ prefixing it, please investigate, confirm the logic with Martin Hillyard and fix if required following the discussion. 

**Investigation**

There was a coding error that inserted the Boolean value. However, the string itself was constructed from the discontinue's pertinent annotation, code and availability time. After discussion with Martin, it was decided a better approach was to use the pertinent information or a default string of "No information available", as per the GP Connect spec.

*Additional changes were identified as a result of the investigation:*

- If a discontinue request is present and it's possible to map the status reason extension, the status of the medication request should be `stopped`.  
- If the status changed date is unknown, then the status should be `completed` and the status reason extension omitted.

### DIFF
--- a/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/medication/MedicationRequestPlanMapper.java
+++ b/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/medication/MedicationRequestPlanMapper.java
@@ -1,6 +1,7 @@
 package uk.nhs.adaptors.pss.translator.mapper.medication;
 
 import static org.hl7.fhir.dstu3.model.MedicationRequest.MedicationRequestIntent.PLAN;
+import static org.hl7.fhir.dstu3.model.MedicationRequest.MedicationRequestStatus.COMPLETED;
 import static org.hl7.fhir.dstu3.model.MedicationRequest.MedicationRequestStatus.STOPPED;
 
 import static uk.nhs.adaptors.pss.translator.mapper.medication.MedicationMapperUtils.buildDispenseRequestPeriodEnd;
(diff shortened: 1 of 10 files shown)


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/medication/MedicationRequestPlanMapper.java
```diff
@@ -224,12 +223,21 @@ private Optional<Reference> extractPriorPrescription(RCMRMT030101UK04Authorise s
     private MedicationRequest.MedicationRequestStatus buildMedicationRequestStatus(RCMRMT030101UK04Authorise supplyAuthorise) {
         if (supplyAuthorise.hasStatusCode() && supplyAuthorise.getStatusCode().hasCode()
             && COMPLETE.equals(supplyAuthorise.getStatusCode().getCode())) {
-            return MedicationRequest.MedicationRequestStatus.COMPLETED;
+            return COMPLETED;
```
SRAlexander, line 226: Should this not be MedicationRequest.MedicationRequestStatus.Completed?

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/medication/MedicationRequestPlanMapper.java
```diff
@@ -224,12 +223,21 @@ private Optional<Reference> extractPriorPrescription(RCMRMT030101UK04Authorise s
...
         } else {
             return MedicationRequest.MedicationRequestStatus.ACTIVE;
```
SRAlexander, line 228: Or simplify this line.

SRAlexander (changes requested):
Tiny change for consistency required but looks good. :)

SRAlexander (approved):

PR #233, opened 2023-02-01
NIAD-2489

MartinWheelerMT:
Subsequent PR for #231 

A resource with one document reference resource, the content field of which is shown below:

```
 "content": [
              {
                "attachment": {
                  "contentType": "text/plain",
                  "url": "https://<url-to-file>",
                  "size": 604,
                  "title": "https://<url-to-file>"
                }
              }
            ],
```
As expected, the attachment.size and attachment.url fields are both present. However, the attachment.title field is also present. The [documentation ](https://developer.nhs.uk/apis/gpconnect-1-6-0/access_documents_development_documentreference.html#contentattachmenttitle)states the title is intended to hold the reason a file isn’t available. Therefore it is not needed here.


SRAlexander (approved):

SRAlexander:
I'm aware of the discussion around the title field, I can see that it has been removed in our use cases. I'm happy for a merge as soon as the build is green.

PR #237, opened 2023-02-14
NIAD-2588: xhl7uk pimp comments

benhession:
### Description

There processing of text/x-hl7uk-pmip comments are not being handled quite as they should be.

Example Comment…

```
<text mediaType="text/x-h7uk-pmip">CommentType:USER COMMENT CommentDate:20100120162723  (EMISTest) - Normal - No Action </text>
```

There are three CommentTypes: `USER COMMENT`, `AGGREGATE COMMENT SET` and `LABORATORY RESULT COMMENT(E141)`

User Comment: Should be a comment against the parent Observation. This appears to be working as expected in the PS Adaptor

Aggregate Comment Set: Should be a comment against the parent Observation. At the moment only the last line of the comment text is being added to the parent comment which needs to be fixed. The full comment text is being added to another observation statement that does not link or is referenced to any other entry, this needs to be removed.

LABORATORY RESULT: This should be added to the linked DiagnosticReport conclusion field. At the moment this is being added to the parent observation, it needs to be moved to the related Diagnostic report with the field name Conclusion.  

### Additional Comment

While investigating the implementation details an additional comment type was found; `LAB SPECIMEN COMMENT(E271)`. Where only the last line was extracted and mapped to the note field of the parent Specimen. An observation (filing note) was also created that was not referenced elsewhere in the ASR.

In order to bring specimen comments in line with the EMIS example, the full comment is now extracted and the surplus observation (filing note) has been removed.


benhession:
#### Additional bug related to comments fixed as part of this ticket.

In the case of observations being direct children of a battery, the mapping has been changed so the narrative statement text is only mapped to the test group observation (mapped from the battery itself) and not the test result observations (mapped from the observation statements).

MartinWheelerMT (approved):

PR #241, opened 2023-02-27
NIAD-2622 - PS Adaptor - (Problems Dataset) authoredOnDate should be set by availability Time

MartinWheelerMT:
As the title suggests ProcedureRequest MedicationRequest authoredOnDate field is incorrect on all counts. It should be pulled from availabilityTime.


benhession (approved):

benhession (commented):
It says in the description the assertedDate should be pulled from availabilityTime . Which availability time is it referring to?

MartinWheelerMT:
I've taken it from the compoundStatement which is an allergyIntolerance, using the ehrComposition and the ehrExtract as fallbacks.

PR #251, opened 2023-03-10
NIAD-2647: Comparator changes

dfk97:


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/QuantityMapper.java
```diff
@@ -78,4 +90,13 @@ private void setQuantityValueAndUnit(Quantity quantity, String value, String uni
         }
         quantity.setValue(new BigDecimal((value)).setScale(decimalPlaceCount, RoundingMode.CEILING));
     }
+
+    private void setQuantityWithNoComparator(Quantity quantity, String value) {
```
benhession, line 94: Should we be adding the unit here too?

benhession (approved):

benhession (approved):

PR #253, opened 2023-03-13
NIAD-2617: bugfix - allergy intolerances secondary codes

benhession:
### Original bug

In the event that we identify an allergy with the snomed code 416098002 then we should look if a value field also exists.

IF it does not then we use the existing functionality.
IF it does then we need to search the xml for the <value> tag and use the coded data from there,

### Investigation

Investigation showed the logic for this was already present in the adaptor. However, because the TPP EHR Extract used a defined namespace (xsi) in the value element the CD wasn’t unmarshalled as expected.  

An issue was also discovered unmarshalling the translation elements when an original text element was the first child.


schema/src/main/java/org/hl7/v3/ValueAdapter.java
```diff
@@ -22,7 +22,7 @@ public class ValueAdapter extends XmlAdapter<Object, Object> {
 
     @Override
     public Object unmarshal(Object v) {
-        if (v instanceof PQ || v instanceof IVLPQ || v instanceof CV) {
+        if (v instanceof PQ || v instanceof IVLPQ || v instanceof CD) {
```
SRAlexander, line 25: This looks fair but are we sure we will never get a CV instance in the unmarshal?

benhession: `instanceof CD` covers CV, as CV is a subclass.

SRAlexander (dismissed):

SRAlexander (approved):

PR #268, opened 2023-03-31
NIAD-2659 - Bug Fix: INT_11: Duplicate NHS number shouldn't be accepted.

MartinWheelerMT:
GIVEN a request is received where the Patient's NHS number is registered against an existing conversation ID request that does not match made request's Conversation ID.
THEN the Facade should return a 500 response with an operationOutcome…

operationErrorCode = "INTERNAL_SERVER_ERROR";
operationErrorMessage = "PS - The Given NHS number is already being processed against Conversation ID: {conversationId}, you cannot start a new request until the current request has completed or failed. ";


MartinWheelerMT:
The large amount of files changed are mainly due to an enum being moved from the facade to common.  This was due to the enum members being required to be accessible in the gp2gp translator

benhession (approved):

PR #270, opened 2023-04-04
NIAD-2678: Bug-fix - Conversation ID is not reset

benhession:
**Description**

The MDC Service stores the conversation ID and it is not being reset after a message is processed. So if messages with different IDs are received, for the second message, the log messages printed before the conversation ID is updated will contain the ID of the first.

The MDC Service needs to be cleared after a message is processed, so log messages printed before the ID is set will have a blank conversation ID.


dfk97 (approved):
LGTM!

PR #277, opened 2023-04-19
COPC / attachment merge error handing updates

benhession:
- Correct Reassembly failure handling. 

- Send COPC ACK after attempting to merge.


benhession:
The COPC needs to be ACK'd before creating the bundle to ensure MIGRATION_COMPLETE is the most recent migration status

SRAlexander (approved):

PR #278, opened 2023-04-27
NIAD-2676 - New monolith Snomed version needs testing.

MartinWheelerMT:
Update Jenkinsfile to use latest SNOMED Monolith version Update test outputs to reflect new mappings


SRAlexander (approved):
Looks like some new code / description matches have been found which allows us to maintain a description code as well as the concept code now. 

No major concerns identified.

PR #283, opened 2023-05-16
NIAD-2707: bugfix - handle MHS Adaptor decoding

benhession:
### Description

The MHS Adaptor is decoding certain file types, examples of these are CSV and XML.

The PS Adaptor is not handling these decoded files correctly.  Transfers that contain XML or CSV files as **non-large attachments** are failed, due to the file length check . As the file length is provided in the ebxml description and represents the length of the base64 encoded file.

We should also test to ensure **large attachments** of the affected file types succeed.

<br/>

**GIVEN** a transfer containing a CSV or XML attachment

**WHEN** the attachment is a non-large message

**THEN** the transfer should be successful and the file should be retrievable from storage. 

<br/>

**GIVEN** a transfer containing CSV or XML attachment

**WHEN** the attachment is a large message

**THEN** the transfer should be successful and the file should be retrievable from storage.


MartinWheelerMT (approved):
Looks good!

PR #284, opened 2023-05-22
NIAD-2713: Bugfix - bundle mapping exception should send NACK code 21

benhession:
If bundle mapping fails after merging attachments, NACK EHR with code 21


benhession:
If bundle mapping fails due to a BundleMappingException after an attachment merge, the adaptor is currently sending NACK code 31 (attachment not received or could not be processed). However, because the exception is thrown when mapping the EHR Extract payload an error code of 21 (EHR Extract message not well-formed or not able to be processed) is more accurate.

PR #288, opened 2023-06-09
Niad-2721: update readme for postman tests

benhession:
Update the readme

### DIFF
--- a/test-suite/README.md
+++ b/test-suite/README.md
@@ -100,4 +100,4 @@ GPC_FACADE_USER_DB_PASSWORD: "123456"
 - Check that both the `ps_gp2gp_translator-1` + `gpc_facade-1` have stopped running
 - Check the certificates in postman
 - If the start script doesn't run, check the file permissions to ensure the script is executable or run with 
-elevated privatises i.e `sudo ./start-test-environment.sh`   
+elevated privileges i.e. `sudo ./start-test-environment.sh`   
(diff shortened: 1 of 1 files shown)


HarveySembhyNHS (approved):
Approved

HarveySembhyNHS (approved):

test-suite/README.md
```diff
@@ -16,11 +16,10 @@
     9. Close Settings menu
 
 ## Setting up test-suite adaptors to run Postman tests
-1. Open project in IDE
-2. Use the terminal and CD into folder `\nia-patient-switching-standard-adaptor\test-suite`
-3. Run the code `sudo ./start-test-environment.sh`
-4. Allow the system to build adaptors
-5. After building your docker suite should look something like this (containers can appear in any order):
+1. Use the terminal and CD into folder `\nia-patient-switching-standard-adaptor\test-suite`
+2. Run the code `sudo ./start-test-environment.sh`
```
dfk97, line 20: You mention to run the code but "sudo" is Linux specific. Should we be specificying that in the instructions?

benhession: I think `sudo` is only necessary when the file isn't flagged as executable on the system. I've moved it to the troubleshooting section. It should always work because its a bash script, so even on windows we run it using WSL and mac is now zsh which is an extension of bash

dfk97 (approved):

PR #296, opened 2023-08-03
NIAD-2793-Patient Switching Assurance - Run 2 - Observation with unit "s" is translated to unit "Siemens"

MartinWheelerMT:
Change map type to HashMap in MeasurementUnitsUtil 
Add unit test for mapped values differing only by case


gp2gp-translator/src/test/resources/xml/Quantity/pq_arbitrary_unit_uppercase_s.xml
```diff
@@ -0,0 +1,5 @@
+<value xmlns="urn:hl7-org:v3" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="urn:hl7-org:v3 ..\Schemas\RCMR_IN030000UK06.xsd" xsi:type="PQ" value="10" unit="1">
```
adrianclay, line 1: Not feedback on your PR, but having this level of indirection, where your test data is outside of the test file really grinds my gears.  Especially when the test fixture is so small, like in this case.  Just put the XML into the Test file!

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/util/MeasurementUnitsUtil.java
```diff
@@ -4,13 +4,13 @@
...
 import javax.annotation.PostConstruct;
+import java.util.HashMap;
 import java.util.Map;
-import java.util.TreeMap;
 
 @Component
 public final class MeasurementUnitsUtil {
 
-    private static final Map<String, String> UNITS = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+    private static final Map<String, String> UNITS = new HashMap<>();
```
adrianclay, line 13: I don't know enough about Java to understand `TreeMap` vs `HashMap`.

adrianclay (approved):

PR #303, opened 2023-08-14
Niad-2810: Bugfix - condition missing code

benhession:
When mapping `Condition`. The adaptor is not currently finding the `code` element in all instanaces. This is because `code` is mapped from the referenced actual problem. In the failing case, the actual problem is an `AllergyIntolerance`, the adaptor is not currently expecting this structure as it contains a compound statement with a nested observation statement component.

The fix was to re-write the `getObservationStatementById` method to check all Observation Statements in the EHR Extract, regardless whether they are nested in an Allergy Intolerance or not.


gp2gp-translator/src/integrationTest/java/uk/nhs/adaptors/pss/translator/mapper/ConditionMapperIT.java
```diff
@@ -0,0 +1,120 @@
...
+
+        final RCMRMT030101UK04EhrExtract ehrExtract = unmarshallEhrExtract(filename);
+        final List<Condition> conditions = conditionMapper.mapResources(ehrExtract, patient, List.of(), PRACTISE_CODE);
+
+        conditionMapper.addReferences(buildBundleWithNamedStatementObservation(), conditions, ehrExtract);
+
+        var code = conditions.get(0).getCode().getCodingFirstRep();
+
+        assertThat(code.getDisplay()).isEqualTo(OBSERVATION_CONCEPT_DESCRIPTION);
+        assertThat(code.getCode()).isEqualTo(OBSERVATION_CONCEPT_ID);
```
oleksiiRybak, line 74: It might be a matter of personal preference but somehow I found "assertEquals(exp, actValue)" more concise and better to read hence I would suggest using that.

benhession: I've updated it to use the JUnit method you suggest. I know it's more verbose, but I like the assertJ `isEqualTo` because it states in the method names where the actual and expected values go. I find the ordering in `assertEquals` a little unintuitive. My brain wants to write `assertEquals(actual, expected)`

oleksiiRybak (dismissed):

oleksiiRybak (approved):

PR #306, opened 2023-08-16
NIAD-2777 - losing local egton codes

MartinWheelerMT:


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/CodeableConceptMapper.java
```diff
@@ -196,6 +204,31 @@ private CodeableConcept generateCodeableConceptUsingConceptId(CD mainCode, CD tr
...
+
+        if (!translationCodes.isEmpty()) {
+            codeableConcept.getCoding().addAll(translationCodes);
+        }
+    }
+
+    private Coding mapNonSnomedCodes(CD codedData) {
+        return new Coding()
+                .setCode(codedData.getCode())
+                .setSystem(CodeSystemsUtil.getFhirCodeSystem(codedData.getCodeSystem()))
```
adrianclay, line 228: `CodeSystemsUtil.getFhirCodeSystem` appears to return the parameter if it isn't aware of the system. Is that behaviour fine for this, or should we be doing something else?

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AllergyIntoleranceMapper.java
```diff
@@ -120,9 +120,13 @@ private void buildCode(AllergyIntolerance allergyIntolerance, RCMRMT030101UK04Co
         if (NON_DRUG_ALLERGY_CODE.equals(compoundCode)) {
             allergyIntolerance.setCode(codeableConceptFromCode);
 
-            if (allergyIntolerance.getCode() != null && !allergyIntolerance.getCode().hasCoding()) {
-                allergyIntolerance.getCode().setCoding(List.of(DegradedCodeableConcepts.DEGRADED_NON_DRUG_ALLERGY));
+            if (allergyIntolerance.getCode() == null) {
+                return;
             }
+
+            DegradedCodeableConcepts.addDegradedEntry(
```
adrianclay, line 127: I got confused by this code to begin with, but I think the problem is the naming of this method `addDegradedEntry` is more along the lines of `addDegradedEntryIfRequired`?

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/CodeableConceptMapper.java
MartinWheelerMT: Yes thats correct, if we don't know the FHIR code system, then we should return the HL7 code system, discussed with Emile in Slack

`Thank you , just read the doc - looks exactly right. The only question is what happens if you cannot translate the HL7 XML Code System (2.16.840.1.113883.2.1.6.3) to a FHIR code system (e.g. https://fhir.hl7.org.uk/Id/egton-codes)? Would you just output the raw value like 2.16.840.1.113883.2.1.6.3`


`
Yes that is correct, if we cannot map a code known code system, then we will output the Raw value. 
`

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AllergyIntoleranceMapper.java
MartinWheelerMT: Yes might make more sense, I'll change this and the other similar named method

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/util/DegradedCodeableConcepts.java
```diff
@@ -40,4 +43,20 @@ public final class DegradedCodeableConcepts {
             .setSystem(SNOMED_SYSTEM)
             .setCode("196411000000103")
             .setDisplay("Transfer-degraded record entry");
+
+    public static void addDegradedEntry(CodeableConcept codeableConcept, Coding degradedCoding) {
```
adrianclay, line 47: 👀 No unit test?

gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/mapper/CodeableConceptMapperTest.java
```diff
@@ -395,7 +397,7 @@ public void mapNoSnomedCodeWithOriginalText() {
         CodeableConcept codeableConcept = codeableConceptMapper.mapToCodeableConcept(codedData);
 
         assertThat(codeableConcept.getText()).isEqualTo("Inactive Problem, minor");
-        assertThat(codeableConcept.hasCoding()).isFalse();
+        assertThat(codeableConcept.getCoding()).hasSize(EXPECTED_CODING_SIZE);
```
adrianclay, line 400: I _hate_ with a passion the magic numbers code style thing. I feel like this test is more confusing now.

gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/mapper/CodeableConceptMapperTest.java
```diff
@@ -468,8 +470,202 @@ public void mapSnomedCodeInTranslationWithDisplayNameWithOriginalText() {
...
+                <code xmlns="urn:hl7-org:v3" code="ALLERGY138185NEMIS" codeSystem="2.16.840.1.113883.2.1.6.3"
+                displayName="Adverse reaction to Comirnaty Covid-19 mRna Vaccine">
+                    <originalText>Test original text</originalText>
+                </code>
+                """;
+
+        var codedData = unmarshallCodeElementFromXMLString(inputXML);
+        var codeableConcept = codeableConceptMapper.mapToCodeableConcept(codedData);
+
+        assertThat(codeableConcept.getText()).isEqualTo("Test original text");
```
adrianclay, line 486: I feel like these asserts might be clearer if they were using the same pattern as the XML in the setup, lines (475-481). That way you could clearly see both the XML and the JSON it would convert to.

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/CodeableConceptMapper.java
adrianclay: Fair enough. I don't have an opinion on what the business requirements should be, just felt a bit strange at first.

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/util/DegradedCodeableConcepts.java
MartinWheelerMT: Oops! Have added them now

gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/mapper/CodeableConceptMapperTest.java
MartinWheelerMT: Will address in a later PR

adrianclay (approved):

PR #311, opened 2023-09-05
NIAD-2817: Degrade codes are not mapped for all resources

MartinWheelerMT:
If a coding block does not contain a Snomed entry (but may contain one or more non-Snomed codes) it should generate a degraded resource.

These changes add the respective degraded code to all resources which map codeableConcepts using the codeableConceptMapper.

Details of degraded codes are detailed in section 3 of the following document:
[Guidance On CodeableConcept](https://developer.nhs.uk/apis/gpconnect-1-6-0/pages/accessrecord_structured/GuidanceOnCodeableConcept.pdf)

### DIFF
--- a/gp2gp-translator/src/integrationTest/resources/e2e-mapping/output-json/PWTP7_vis-output.json
+++ b/gp2gp-translator/src/integrationTest/resources/e2e-mapping/output-json/PWTP7_vis-output.json
@@ -3493,15 +3493,6 @@
             ]
           }
         ],
-        "code": {
-          "coding": [
-            {
-              "system": "http://snomed.info/sct",
-              "code": "196411000000103",
-              "display": "Transfer-degraded record entry"
-            }
-          ]
-        },
         "subject": {
           "reference": "Patient/b362eb8b-4952-4daa-83fe-c86bbd3f1f2e"
         },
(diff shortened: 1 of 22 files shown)


MartinWheelerMT:
This pull request has a large number of files, however each commit is specifically for one mapping type and it will be probably be easier to review by looking at the changes in each commit

gp2gp-translator/src/integrationTest/resources/e2e-mapping/output-json/PWTP7_vis-output.json
```diff
@@ -3493,6 +3493,15 @@
...
         ],
+        "code": {
+          "coding": [
+            {
+              "system": "http://snomed.info/sct",
+              "code": "196411000000103",
+              "display": "Transfer-degraded record entry"
+            }
+          ]
+        },
```
adrianclay, line 3504: I'm a bit confused as to why we've now got a value for `code` while not having a value previously. 🤔

MartinWheelerMT: After discussion, have moved the logic into a different section to prevent a degraded coding being added if there is no code in the codeable concept

adrianclay (dismissed):

adrianclay (approved):

PR #320, opened 2023-09-28
Enable AWS instance roles for S3

benhession:
## What

Enable S3 connections when using an AWS instance role. 

## Why

The current implementation always requires user credentials. The change allows credentials to be used when they are provided, otherwise attempts a connection without. This makes it possible for authorisation to be configured via an AWS Instance role.     

## Type of change

## Checklist:

- [x] I have updated the [Changelog](CHANGELOG.md) with details of my change in the UNRELEASED section if this change will affect end users

### DIFF
--- a/developer-information.md
+++ b/developer-information.md
@@ -50,6 +50,11 @@ Please make sure to load the latest release of Snomed CT UK Edition. See [Config
     - KEY_PASSWORD: server private key password
     - TRUST_STORE: path to the truststore
     - TRUST_STORE_PASSWORD: truststore password
+    - STORAGE_TYPE: the type of object storage to use for attachments (S3, Azure or LocalMock)
+    - STORAGE_REGION: The AWS region (leave undefined if using an AWS instance role)
+    - CONTAINER_NAME: The name of the Azure Storage container or Amazon S3 Bucket
+    - STORAGE_REFERENCE: The Azure account name or AWS Access Key ID (leave undefined if using an AWS instance role)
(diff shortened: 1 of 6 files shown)


developer-information.md
```diff
@@ -50,6 +50,11 @@ Please make sure to load the latest release of Snomed CT UK Edition. See [Config
     - KEY_PASSWORD: server private key password
     - TRUST_STORE: path to the truststore
     - TRUST_STORE_PASSWORD: truststore password
+    - STORAGE_TYPE: the type of object storage to use for attachments (S3, Azure or LocalMock)
+    - STORAGE_REGION: The AWS region (leave undefined if using an AWS instance role)
+    - CONTAINER_NAME: The name of the Azure Storage container or Amazon S3 Bucket
+    - STORAGE_REFERENCE: The Azure account name or AWS Access Key ID (leave undefined if using an AWS instance role)
+    - STORAGE_SECRET: The Azure account key or the access key for Amazon S3. (leave undefined if using an AWS instance role)
```
adrianclay, line 57: I personally find it a bit strange that the PS Adaptor has these environment variables.

I'd think it would make sense to use the default behaviour of the respective Java SDKs and link off to both the [AWS](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials-chain.html#credentials-default) and [Azure](https://learn.microsoft.com/en-us/azure/developer/java/sdk/identity-azure-hosted-auth#default-azure-credential) guidance for configuring access.

adrianclay (approved):

PR #324, opened 2023-10-04
NIAD-2853: Fixed issue with coding text being set from incorrect place when mapping allergy intolerances

MartinWheelerMT:
## What

This code change is to make the allergy intolerance mapper use the `displayName` of the <value> block when setting the value for `code.text`.
It also fixes some bad data in the related test data for the EhrExtractHandlingIT tests and introduces an additional unit test.

## Why

This issue was raised by under NIAD-2853 and will provide more relevant content when the Json Bundle is used by the consumer

## Type of change

## Checklist:

- [x] I have updated the [Changelog](CHANGELOG.md) with details of my change in the UNRELEASED section if this change will affect end users


adrianclay:
> NIAD-2853: Unexpected Allergy code text displayed in Medicus

Very confusing title, I think Medicus is only incidentally involved here.

I much prefer the text in the CHANGELOG.md "Fixed issue with coding text being set from incorrect place when mapping allergy intolerances."

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AllergyIntoleranceMapper.java
```diff
@@ -190,6 +190,13 @@ private DateTimeType getAssertedDateElement(TS availabilityTime, RCMRMT030101UK0
     private void buildAllergyIntoleranceText(RCMRMT030101UK04ObservationStatement observationStatement,
```
adrianclay, line 190: Do the tests pass if we remove this method entirely? I'm confused as to why we need to manipulate the code text at all.

gp2gp-translator/src/test/resources/xml/AllergyIntolerance/allergy-structure-with-original-text-in-code.xml
```diff
@@ -29,9 +29,6 @@
                                     <translation code="416098002" codeSystem="2.16.840.1.113883.2.1.3.2.4.15" displayName="Allergy to drug"/>
                                     <translation code="14L..00" codeSystem="2.16.840.1.113883.2.1.6.2" displayName="H/O: drug allergy"/>
                                 </code>
-                                <value type="CD" codeSystem="1.3.6.1.4.1.17340.1.5" code="9006113"
-                                       displayName="Propranolol 10mg tablets (Almus Pharmaceuticals Ltd) @ 28 tablet(s)">
-                                </value>
```
adrianclay, line 34: We've modified this test fixture, and I'm confused as to whether it's a valuable test at all anymore? Would we ever expect to see a `CompoundStatement` which looked like this?

It feels a little to me like we're creating a test fixture to match some behaviour our code has, as opposed to the other way round.

adrianclay: Was NIAD-2621 the original ticket which prompted this behaviour/test? Perhaps we should go back to the `TPP` GP2GP XML extract as a source and see what it looked like.

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AllergyIntoleranceMapper.java
MartinWheelerMT: It looks like the codeable concept mapper does not map the text field as is mapped using a conceptId and no original text has been entered.  In this case the text value is set to null.

I've tried multiple ways of changing this in the codeable concept mapper, but results in a lot of changes to the output for other elements which use codeable concept mapper.

For allergyIntolerance, this issue requires this text value to be set, so unfortunately I believe we need to leave this in.

I have refactored the buildCode method in the allergyIntolerance mapper to make it clear to understand and remove some nesting

gp2gp-translator/src/test/resources/xml/AllergyIntolerance/allergy-structure-with-original-text-in-code.xml
MartinWheelerMT: Whilst on the Medicus call, there was an example provided which had no value but had an originalText in the code section, so I would suspect this is a valid use case.

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AllergyIntoleranceMapper.java
```diff
@@ -124,39 +150,45 @@ private void buildCode(AllergyIntolerance allergyIntolerance, RCMRMT030101UK04Co
...
+            }
+
+            allergyIntolerance.setCode(codeableConceptFromValue);
+
+            var allergyNote = getAllergyNote(codeableConceptFromCode, codeableConceptFromValue);
+            if (allergyNote != null) {
+                allergyIntolerance.getNote().add(allergyNote);
+            }
+        } else {
+            allergyIntolerance.setCode(codeableConceptFromCode);
```
benhession, line 180: From my understanding `mapToAllergyIntollerence` wouldn't be called if the compound code was not  `SN53.00` or `14L..00`  so we could throw an exception here (or do nothing).

benhession (commented):
Does the [mapping documentation](https://github.com/NHSDigital/patient-switching-adaptors-mapping-documentation/tree/main/allergy%20intolerances) need updating with this change?

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AllergyIntoleranceMapper.java
MartinWheelerMT: This seems be correct, I don't believe we need this final else

benhession (approved):

PR #328, opened 2023-10-12
NIAD-2777: Bug Fix - attachments not decompressed

benhession:
## What

Update the code that builds the file description when an attachments metadata is re-built from the database.  Refactor the code to reduce duplication of this functionality.

Enable logging for migration status updates using `PS_LOGGING_LEVEL`.

## Why

Compressed large attachments are not decompressed when uploading to object storage. This is due to the description being re-build incorrectly i.e. `Compressed=true` instead of  `Compressed=Yes`.  The skeleton attribute was also re-built incorrectly as `DomainData=”\X-GP2GP-Skeleton:Yes\”` where the documentation states it should be `DomainData=”X-GP2GP-Skeleton: Yes`.

SA testing have requested the migration status updates to be printed to the logs. This was already present but not configurable via the environment variables. 

## Type of change

## Checklist:

### DIFF
--- a/db-connector/src/main/java/uk/nhs/adaptors/connector/model/PatientAttachmentLog.java
+++ b/db-connector/src/main/java/uk/nhs/adaptors/connector/model/PatientAttachmentLog.java
@@ -28,15 +28,13 @@ public class PatientAttachmentLog {
     private Boolean isBase64;
 
     public String getFileDescription() {
-        String description = String.format(
-            "Filename=\"%s\" ContentType=%s Compressed=%s LargeAttachment=%s OriginalBase64=%s Length=%s",
-            getFilename(),
-            getContentType(),
-            getYesNoString(getCompressed()),
-            getYesNoString(getLargeAttachment()),
-            getYesNoString(getOriginalBase64()),
-            getLengthNum()
-        );
+        String description =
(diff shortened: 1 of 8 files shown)


db-connector/src/main/java/uk/nhs/adaptors/connector/model/PatientAttachmentLog.java
```diff
@@ -22,10 +21,31 @@ public class PatientAttachmentLog {
...
     private Boolean uploaded;
     private Integer lengthNum;
-//  @NonNull patientMigrationReqId required to create
     private Integer patientMigrationReqId;
     private Integer orderNum;
     private Boolean deleted;
     private Integer postProcessedLengthNum;
     private Boolean isBase64;
+
+    public String getFileDescription() {
```
ole4ryb, line 30: I am just wondering if we are expecting a lot of requests of this method. If so then maybe concatenation is a better option than String format as it works faster ?

benhession: Potentially yes. I've changed it to use concatenation.

ole4ryb (approved):

PR #329, opened 2023-10-13
NIAD-2866: Bug Fix - Final ACK not accepted by TPP

benhession:
## What

Update ACK template with correct namespaces and class codes. 

## Why

The final Ack is not accepted by TPP

## Type of change

## Checklist:

- [-] I have made corresponding changes to the documentation
- [-] I have added tests that prove my fix is effective or that my feature works


adrianclay (approved):
I'm surprised that no tests failed, but I'm also not confident that adding tests here would be worthwhile. They'd just become very tightly mirrored to the XML.

PR #330, opened 2023-10-13
NIAD-2873: changing the way how we querying SDS service

ole4ryb:
## What

By this PR we are adapting the way how we query SDS service to avoid getting a broken response.

## Why

When querying the SDS service and sending ODS and identifier params we get some sections  of the response empty which  causes some workflow issues. The ITOC team couldn't provide any solutions that could resolve the issue. There are some alternative endpoints that we can use and find the workaround and that's what this PR is bringing.

## Type of change

## Checklist:

### DIFF
--- a/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/sds/SdsRequestBuilder.java
+++ b/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/sds/SdsRequestBuilder.java
@@ -19,21 +19,23 @@ import uk.nhs.adaptors.pss.translator.service.RequestBuilderService;
 public class SdsRequestBuilder {
 
     private static final String ROUTING_AND_READABILITY_ENDPOINT = "/Endpoint";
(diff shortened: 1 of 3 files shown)


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/sds/SdsRequestBuilder.java
```diff
@@ -19,21 +19,23 @@
 public class SdsRequestBuilder {
 
     private static final String ROUTING_AND_READABILITY_ENDPOINT = "/Endpoint";
+    private static final String ROUTING_AND_READABILITY_DEVICE_ENDPOINT = "/Device";
```
benhession, line 22: Because this is the device endpoint I don't think is should be called "Routing and reliability" the [docs](https://digital.nhs.uk/developer/api-catalogue/spine-directory-service-fhir#get-/Device) call it Get Accredited System Details

benhession (commented):
Minor comment around constant name, but otherwise looks good!

PR #345, opened 2023-10-26
NIAD-2889: Bugfix - add additional JMS error handling

benhession:
## What

- Add extra handling and logging to the JMS Listener error handler.
- Remove unused SQL statement logging

## Why

If an exception is thrown with a null cause the handler itself throws an exception. This prevents the JMS message delivery count property from being updated and leads to the message being dequeued repeatedly. 

The null pointer would also prevent any logging of the exception message. Therefore, in addition to handling an unexpected exception without a cause, extra logging has been added.

SQL statement logging had been setup during development but was not configurable to be used by the end user. When saving the inbound message, the logging statement placed the statement, including the entire message on the stack. Testing the service with a very low heap size (128 MB), this logging caused an out of memory error.   

## Type of change

## Checklist:


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/amqp/JmsListenerErrorHandler.java
```diff
@@ -19,11 +19,17 @@ public class JmsListenerErrorHandler implements ErrorHandler {
...
         Throwable cause = t.getCause();
-        Class<? extends Throwable> classOfCause = cause.getClass();
+        if (cause == null) {
+            return;
+        }
 
+        Class<? extends Throwable> classOfCause = cause.getClass();
         LOGGER.warn("Caught Error cause of type: [{}], with message: [{}]", classOfCause.toString(), cause.getMessage());
 
         if (RETRYABLE_EXCEPTION_MESSAGES.containsKey(classOfCause)) {
```
adrianclay, line 35:
```suggestion
        // Rety these specific exceptions continuously until they stop happening.
        // These retries will happen until the associated transfer times out.
        if (RETRYABLE_EXCEPTION_MESSAGES.containsKey(classOfCause)) {
```

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+* add additional error handling for unexpected exceptions
+* modify SQL statements to guard against out of memory errors
```
adrianclay, line 9:
```suggestion
### Fixed

* Add additional error handling for exceptions raised when processing PSS queue, and MHS queue messages.
* Fix bug in some SQL statement which was causing excessively large amounts of data was being returned, causing PostgresSQL Out of Memory errors.
```

adrianclay (approved):

PR #348, opened 2023-10-30
NIAD-2902: Codeable concept ordering

MartinWheelerMT:
## What

The `CodeableConceptMapper` has been changed so that the original ordering is preserved.
A new unit test has been added to confirm ordering and all other unit tests / integration tests affected by this change have been updated.

## Why

Initial reported as an issue with the ordering of codes in `Blood Pressures` where the code which was wanted to be received as the first element in the coding array, was appearing in a different order in the original XML source.   This is because the SNOMED code was set to always be the first element (in the original requirements we would only map SNOMED codes, not other codes sets).   

This led to a further discussion where it was decided that when mapping the codes, we should always preserve the original order.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

This is a breaking change as the ordering of all code elements has the potential to be different between versions.

## Checklist:


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/CodeableConceptMapper.java
```diff
@@ -205,28 +206,23 @@ private CodeableConcept generateCodeableConceptUsingConceptId(CD mainCode, CD tr
     }
 
     private void addNonSnomedCodesToCodeableConcept(CodeableConcept codeableConcept, CD codedData) {
```
adrianclay, line 208: I got a bit confused looking at this code, but I think my confusion stems from the name.

This method appears to populate the Coding block for all codes, snomed and non-snomed. Am I correct in that understanding? Perhaps the method should be called something like `populateCodableConceptCodings`?

MartinWheelerMT: I'm happy to take a look at refactoring it, it is pretty complicated in the way the snomed code / extension is built, but I'm sure it can be simpler

adrianclay (approved):

PR #353, opened 2023-11-01
Making util classes effectively sealed by restricting the constructor…

ole4ryb:
Util classes provide the functionality methods but haven't been sealed properly.

## What
Constructor restriction has been added to make the util classes properly sealed.

## Why

Util classes provide the functionality methods but haven't been sealed properly so it is possible to create unnecessary classes instances which creates ambiguity about the intentional usage of such classes.

## Type of change

## Checklist:

### DIFF
--- a/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/util/CodeSystemsUtil.java
+++ b/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/util/CodeSystemsUtil.java
@@ -2,8 +2,6 @@ package uk.nhs.adaptors.pss.translator.util;
 
 import lombok.AccessLevel;
 import lombok.NoArgsConstructor;
-import org.springframework.stereotype.Component;
-
 import java.util.Map;
 
 @NoArgsConstructor(access = AccessLevel.PRIVATE)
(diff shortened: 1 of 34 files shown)


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/util/CodeSystemsUtil.java
```diff
@@ -1,10 +1,13 @@
 package uk.nhs.adaptors.pss.translator.util;
 
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
 import org.springframework.stereotype.Component;
 
 import java.util.Map;
 
 @Component
```
benhession, line 9: Can the `@Component` annotation be removed? It doesn't seem to be injected anywhere.

benhession: Also, the `@Component` annotation identifies the class as a bean. Without a no argument public constructor it technically isn't. So I think we should remove it.

ole4ryb: Yes, agree, it is not used as a Spring component.

benhession (approved):

PR #361, opened 2023-11-06
Remove the ARM specific scripts within docker folder

adrianclay:
## What

I could see no reason for the ARM specific files apart from the activemq image.

The base image for the activemq image didn't have an ARM image available. Furthermore the [base image had been deprecated](https://adoptopenjdk.net/upstream.html) by RedHat the maintainers.

Switching to eclipse-temurin will give us a up to date base image, with ARM support.

Note that only the 21-jdk-alpine version supports ARM, 17 does not.

Also expand out the arguments to `adduser` and `addgroup` to be more readable.

## Why

Simplifying these scripts should reduce confusion/difficulties when developing on different machines.

## Type of change

## Checklist:


MartinWheelerMT (approved):
Tested locally and environment builds, tests and runs fine

PR #363, opened 2023-11-07
NIAD-2858: show nack error code

ole4ryb:
## What

Please include a summary of the changes and the related issue

## Why

Please include details of the reasoning for these changes

## Type of change

## Checklist:


common/src/main/java/uk/nhs/adaptors/common/util/CodeableConceptUtils.java
```diff
@@ -29,6 +30,16 @@ public static CodeableConcept createCodeableConcept(String code, String system,
         return codeableConcept;
     }
 
+    public static CodeableConcept createCodeableConceptWithDoubleCoding(String code, String system, String display,
```
MartinWheelerMT, line 33: This should be a more specific name for the method, as it specifically creates a codeable concept with the oid for "EhrRequestAckCode", whereas the name suggests it is more generic

common/src/main/java/uk/nhs/adaptors/common/util/CodeableConceptUtils.java
```diff
@@ -29,6 +30,16 @@ public static CodeableConcept createCodeableConcept(String code, String system,
...
+                                                                        String text, String gp2gpSpecificCode) {
+        var codeableConcept = new CodeableConcept().setText(text);
+        var coding = new Coding(system, code, display);
+        var codingWithGP2GPSpecificCode = new Coding("2.16.840.1.113883.2.1.3.2.4.17.101", gp2gpSpecificCode, display);
```
MartinWheelerMT, line 37: Maybe pull the oid out into a const rather than hard coding in the method, something like ACKNOWLEDGEMENT_CODE_SYSTEM

common/src/main/java/uk/nhs/adaptors/common/util/FileUtil.java
```diff
@@ -5,11 +5,15 @@
 import java.io.FileNotFoundException;
 import java.io.InputStream;
 
+import lombok.AccessLevel;
+import lombok.NoArgsConstructor;
 import org.apache.commons.io.IOUtils;
 
 import lombok.SneakyThrows;
 
+@NoArgsConstructor(access = AccessLevel.PRIVATE)
```
MartinWheelerMT, line 14: Just a question here.  What is this line that has been added doing?

MartinWheelerMT (commented):
Left a few comments about method names and consts

common/src/main/java/uk/nhs/adaptors/common/util/FileUtil.java
ole4ryb: it just makes this Util effectively working as a util class and not something that can be instantiated with "new FileUtil" class.  It is just making whatever we declare by name (....Util) to work as it was declared..

MartinWheelerMT: ahh like static abstract class.. thanks

MartinWheelerMT (approved):

PR #365, opened 2023-11-07
NIAD-2797: ReferralRequest priority should not use 'Stat'

MartinWheelerMT:
## What

Updated the Referral Request Mapper to use 'ASAP' instead of 'Stat' when mapping and add unit tests to cover these scenarios.
Have also refactored the tests for Referral Request Mapper and the mapper itself.

## Why

As per the documentation: [Request Priority](https://fhir.nhs.uk/STU3/ValueSet/GPConnect-RequestPriority-1) we should be using 'stat' not 'asap' when mapping referral request priority codes.

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Checklist:


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 * Updated the way immunization codes are loaded to build these from the relationships reference set in the SNOMED CT 
 release rather than the outdated CSV file.
+* Update ReferralRequest mapping to use `ASAP` instead of `Stat`.
```
adrianclay, line 10: I wonder if this is a breaking change 🤣 🤔.

MartinWheelerMT: I can update to report as breaking change!

adrianclay (approved):
Enjoy the `assertAll`, tried playing with it locally and felt somewhat disappointed at verbose output JUnit produces. One day Java will learn what concise means.

PR #371, opened 2023-11-10
NIAD-2772: adding identifier block to practitioner section

ole4ryb:
## What

Practitioner GP Number Field Missing from FHIR Output

## Why

GP Number Field that is included in agentDirectory section doesn't appear in FHIR Output Practitioner section. This PR adds this information into Practitioner section.

## Type of change

## Checklist:

### DIFF
--- a/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AgentDirectoryMapper.java
+++ b/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AgentDirectoryMapper.java
@@ -1,5 +1,6 @@
 package uk.nhs.adaptors.pss.translator.mapper;
 
+import static org.apache.logging.log4j.util.Strings.isNotEmpty;
 import static uk.nhs.adaptors.pss.translator.util.OrganizationUtil.findDuplicateOrganisation;
 import static uk.nhs.adaptors.pss.translator.util.ResourceUtil.generateMeta;
 
@@ -99,7 +100,7 @@ public class AgentDirectoryMapper {
         practitioner.setMeta(generateMeta(PRACT_META_PROFILE));
         practitioner.setName(getPractitionerName(agentPerson.getName()));
 
-        if (gpNumber != null && !"".equals(gpNumber)) {
+        if (isNotEmpty(gpNumber)) {
             Identifier identifier = new Identifier()
                 .setSystem("https://fhir.hl7.org.uk/Id/gmp-number")
                 .setValue(gpNumber);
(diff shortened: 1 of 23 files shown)


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AgentDirectoryMapper.java
```diff
@@ -85,19 +86,26 @@ private void mapAgent(RCCTMT120101UK01Agent agent, List<DomainResource> agentRes
...
 
-    private Practitioner createPractitioner(RCCTMT120101UK01Person agentPerson, String id) {
+    private Practitioner createPractitioner(RCCTMT120101UK01Person agentPerson, String id, String gpNumber) {
         var practitioner = new Practitioner();
 
         practitioner.setId(id);
         practitioner.setMeta(generateMeta(PRACT_META_PROFILE));
         practitioner.setName(getPractitionerName(agentPerson.getName()));
 
+        if (gpNumber != null && !"".equals(gpNumber)) {
```
MartinWheelerMT, line 102: Can use StringUtils.isNotEmpty(gpNumber) here to avoid the messy comparison statement, as that covers null, empty or whitespace strings in it's comparison

ole4ryb: Yes, that's a good point. I will add it.

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AgentDirectoryMapper.java
```diff
@@ -71,9 +71,10 @@ private void mapAgent(RCCTMT120101UK01Agent agent, List<DomainResource> agentRes
         var agentOrganization = agent.getAgentOrganization();
         var representedOrganization = agent.getRepresentedOrganization();
         var resourceId = agent.getId().get(0).getRoot();
+        var gpNumber = agent.getId().size() > 1 ? agent.getId().get(1).getExtension() : "";
```
MartinWheelerMT, line 74: I think using !isEmpty() here is more readable, and i believe is best practice as in some collections .size will have to count all the items before checking, whereas isEmpty() will short-circuit on a single value found

MartinWheelerMT: Also gpNumber can only come from one source, so the string value should really be null not an empty string as there is no intention of adding to it at a later point in the code

ole4ryb: while I agree about the isEmpty, I intentionally don't assign gpNumber null value as it is not a good practice I think and is prone to NPE errors in different places. "" string is more safer and doesn't cause any side effects.

ole4ryb: actually we need to use agent.getId().size() > 1 as in this case I need to specifically check that collection has more than one element stored...

MartinWheelerMT: Oh sorry yest I misread as >= 1 :), thats fine!

MartinWheelerMT (approved):

PR #374, opened 2023-11-17
Updated CHANGELOG.md and release.sh to latest version of adaptor

HarveySembhyNHS:
## What

Changelog and release script

## Why

- Updated changelog and release script to match with version 1.2.0

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,14 +6,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ## [Unreleased]
 
 ### Updated
+*
 
+### Changed
+*
+
+### Fixed
+
+* Changed 'author' to be used as Recorder and 'practitioner' to be used as Asserter in AllergyIntolerance.
+
+## [1.2.0] - 2023-11-17
+
+### Updated
+* Updated Practitioner resource type to include gmp numbers.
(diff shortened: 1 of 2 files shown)


MartinWheelerMT (approved):

CHANGELOG.md
```diff
@@ -5,18 +5,18 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
...
-* Updated Practitioner resource type to include gmp numbers.
-* Updated the way Observation comments are mapped so that blood pressure NarrativeStatements are not duplicated into
-  in a separate observation.
+### Added
+
+* Populate `identifier` field for `Practitioner` resources using the General Practitioner (GMP) number if available.
+
+### Fixed
+
+* Prevent `NarrativeStatement / text` sat within a blood pressure `CompoundStatement` from being mapped
```
adrianclay, line 17: `Updated` is not a [valid type of changelog entry](https://keepachangelog.com/en/1.1.0/#how). #pedant.

But being less pendantic, updated is a very generic term, and could cover all and every type of change made. `Fixed` feels more appropriate given it was a bug, most users won't notice a difference, and no additional functionality has been added/removed that the NME needs to be aware of.

adrianclay (approved):

PR #375, opened 2023-11-17
NIAD-2931: Prepend unique ID to fragment filename when storing inside of the database

benhession:
## What

Prepend unique ID to fragment filename when storing to the database / S3.

## Why

Workaround to fix issue with TPP not sending unique filenames for attachment fragments.

## Type of change

## Checklist:


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/util/XmlParseUtilService.java
```diff
@@ -264,14 +263,11 @@ public static String parseMessageRef(RCMRIN030000UKMessage payload) {
...
     public static String parseFragmentFilename(String description) {
-        try {
-            return Arrays.stream(description.split(" "))
-                    .filter(desc -> desc.contains("Filename"))
-                    .map(desc -> desc.replace("Filename=", "").replace("\"", ""))
-                    .toList().get(0);
-        } catch (IndexOutOfBoundsException e) {
-            return "";
-        }
+        return Arrays.stream(description.split(" "))
```
ole4ryb, line 266: If we are removing try-catch block then are we confident that it will be intercepted and dealt with up the ladder...?

ole4ryb (dismissed):

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/util/XmlParseUtilService.java
adrianclay: I find this version more readable compared to the `try/catch`, but I'm a little unsure about using the description as a filename.

I'm guessing the assumption here is that the description isn't within the form `Filename=""` but as a raw filename. If that is the assumption we're making, this approach makes sense.

Also feels janky to assume that the filename doesn't contain a space itself.  E.g. a description of:
`Filename="Adrians fantastic cake.jpg" Size=100`

benhession: The assumption is if the description doesn't contain `Filename=""` then it we expect a description tag with a raw filename and no other data.

adrianclay (approved):

PR #387, opened 2023-12-12
NIAD-2955: Medication resource missing when Medication Context has not been reset

MartinWheelerMT:
## What

Have introduced a new Unit Test and due to DI Scopes, the best way to fix this appears to be with better exception handling for the medication mapper to ensure the Medication context is reset regardless of succe
## Why
Due to an edge case in the medication mapping that occurs when a EhrExtract has failed processing whilst mapping a medication, and that previous extract had already mapped a medication which is also in the next EhrExtract to be processed.
This leads to this medication not been mapped in the final bundle.

## Type of change

## Checklist:


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/medication/MedicationRequestMapper.java
```diff
@@ -25,7 +25,7 @@
 import org.springframework.stereotype.Service;
 
 import lombok.AllArgsConstructor;
-
+]
```
ole4ryb, line 28: Martin, I think this causes a failure and needs to be removed.

MartinWheelerMT: Yes, that has been removed there, that was left over in error when clearing unused exports / testing scoping.

ole4ryb (approved):

adrianclay:
I wonder if this functionality is better tested at the unit level:
- E2E tests are slow
- If the scenario stops throwing an exception, re-creating this failure within the E2E context will be very difficult.
- Unit tests are good at injecting random exceptions (hopefully)

adrianclay (approved):

PR #425, opened 2024-01-16
Added parameterised unit test to test asserted date is mapped from th…

MartinWheelerMT:
## What

Added parameterised unit test to test asserted date is mapped from the available primary and fallback date fields.
Added test XML files for the above test.
Removed dependency on EhrExtract for AllergyIntoleranceMapper and associated fallback to EhrExtract.AvailabilityTime Added fallback for EhrComposition.Author.Time where available Updated CHANGELOG.md

## Why

EhrExtract / availabilityTime field is described as a date which has no relevance to this field and should not be used for this case:

`System date and time of generating this extract on the system from which the message originates.`

## Type of change

## Checklist:

Mapping Documentation will be updated once all the remapping has been done

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 * REST buffer size has been set to 150Mb
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


MartinWheelerMT:
Awaiting response from Martin Hillyard on NIAD-2935, moving to draft

PR #427, opened 2024-01-19
Niad-2922: making classes more interface-oriented

ole4ryb:
## What

Interfaces were introduced into RCMR_IN030000UK06-related  message objects

## Why

In order to fully accommodate the new RCMR_IN030000UK07 message type interfaces should be introduced to make code more flexible. Also, introducing interfaces follows the best code practises. 

## Type of change

## Checklist:


schema/src/main/java/org/hl7/v3/RCMRMT030101UKAgentDirectory.java
```diff
@@ -0,0 +1,22 @@
+package org.hl7.v3;
```
adrianclay, line 1: Not a blocker, but I'd suggest that these interfaces are more strongly assoccited with the PSS package, as that's what cares about their implementation.

adrianclay (approved):
I haven't had a fine comb through all the changes, more a skim. 😄

schema/src/main/java/org/hl7/v3/RCMRMT030101UKAgentDirectory.java
ole4ryb: ok, I think it is a good call, I will consider it in some later PRs.

PR #455, opened 2024-02-09
NIAD-3003: Performance of translator slowed down because of SNOMED DB indexing

MartinWheelerMT:
## What

The build database scripts have been update to build and refresh a material view to retrieve the preferred terms 
The affected SQL query has been updated to us the new materialise view, now ordering by the active flag to ensure consistent ordering
The remaining queries which use the description reference table have been updated to order by the active flag to ensure consistent ordering

## Why

The SQL query used was quite inefficient.  As the tables are read-only based on data populated database creation, a materialised view seemed the correct approach to address this
The remaining queries which use the description reference table search only on the indexed column so do require additional indexing

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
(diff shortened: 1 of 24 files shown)


adrianclay:
Code looks good to me @MartinWheelerMT though I think we need a big warning in the CHANGELOG to be clear that the user will need to run the SNOMED Import script before they run this new version of the translator.

adrianclay:
Also a bit strange that your commits include one from Dependabot. Some strange `git` shenannigans.

adrianclay (approved):

PR #478, opened 2024-02-20
NIAD-3038 & 3039: Plan Statement (Recall) Statuses

MartinWheelerMT:
## What

Previously the adaptor had a fixed value of `Active` for Plan Statements, this has now been changed so that the following mapping occurs when `PlanStatement.text` starts with `Status:`:

`Status: Pending` -> `active`
`Status: Cancelled by clinician` -> `cancelled`
`Status: Superseded` -> `cancelled`
`Status: Seen` -> `completed`

When `PlanStatement.text` is not present or does not not start with `Status:` then a fallback of `unknown` is used.

## Why

The adaptor has a fixed value of `Status: active` for Plan Statements, but this does not reflect the status being returned by **SystmOne**.   When a recall is generated by **SystmOne** they consistently provide a value for `PlanStatement.text` with starts with 'Status: `; The value immediately following this should be used to set the appropriate status in the resource.

Although this is consistently output by **SystmOne**, **EMIS Web** does not output a `PlanStatement.Text` value at all for Diary Entries.  This means that fallback was necessary for when the `PlanStatement.Text` does not exist or does not contain a status.  
This was covered in [NIAD-3039](https://gpitbjss.atlassian.net/browse/NIAD-3039) and so made sense to implement this at the same time.

## Type of change

## Checklist:

### DIFF
--- a/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ProcedureRequestMapper.java
+++ b/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ProcedureRequestMapper.java
@@ -34,6 +34,11 @@ import uk.nhs.adaptors.pss.translator.util.ParticipantReferenceUtil;
 public class ProcedureRequestMapper extends AbstractMapper<ProcedureRequest> {
     private static final String META_PROFILE = "ProcedureRequest-1";
 
+    private static final String STATUS_PENDING = "status: pending";
+    private static final String STATUS_CLINICIAN_CANCELLED = "status: cancelled by clinician";
+    private static final String STATUS_SUPERSEDED = "status: superseded";
+    private static final String STATUS_SEEN = "status: seen";
+
     private final CodeableConceptMapper codeableConceptMapper;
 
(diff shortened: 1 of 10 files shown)


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ProcedureRequestMapper.java
```diff
@@ -104,6 +109,23 @@ private Annotation getNote(String text) {
         return null;
     }
 
+    private ProcedureRequestStatus getStatus(String planStatementText) {
+        if (planStatementText == null || planStatementText.isEmpty()) {
```
ole4ryb, line 113: I think you can use StringUtils in here which should be more concise option for checking for nulls and emptiness

ole4ryb (approved):

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ProcedureRequestMapper.java
MartinWheelerMT: doh! Thats absolutely right, I'll fix that in future pull request touching this mapper :)

PR #492, opened 2024-02-29
Fix formatting of polling responses

adrianclay:
## Why

Markdown had become corrupt

## Type of change


HarveySembhyNHS (approved):
Approved - formatting looks correct

PR #499, opened 2024-03-04
JMS refactoring and version update

ole4ryb:
## What

JMS version was uplifted. Also, JMSConnectionFactory was refactored and closeTimeout was introduced in an attempt to imrpove JMS client performance. 

## Why

While doing performance tests of the PS Adaptor it was noticed that the adaptor started failing after some threshold of messages was crossed. The failure were related to JmsConnectionFactory hence we should optimizing the config to make the performance of JMS messages more efficient. 

## Type of change

## Checklist:

- [] New and existing unit tests pass locally with my changes

### DIFF
--- a/gp2gp-translator/src/integrationTest/resources/application.yml
+++ b/gp2gp-translator/src/integrationTest/resources/application.yml
@@ -26,7 +26,7 @@ amqp:
     username: ${PS_AMQP_USERNAME:}
     password: ${PS_AMQP_PASSWORD:}
     maxRedeliveries: ${PS_AMQP_MAX_REDELIVERIES:3}
-    closeTimeout: ${PS_CLOSE_TIMEOUT:6000}
+    closeTimeout: ${PS_CLOSE_TIMEOUT:10000}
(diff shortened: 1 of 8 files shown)


gp2gp-translator/src/integrationTest/resources/application.yml
```diff
@@ -26,18 +26,21 @@ amqp:
     username: ${PS_AMQP_USERNAME:}
     password: ${PS_AMQP_PASSWORD:}
     maxRedeliveries: ${PS_AMQP_MAX_REDELIVERIES:3}
+    closeTimeout: ${PS_CLOSE_TIMEOUT:10000}
```
adrianclay, line 29: Are we intentionally settin the timeout less than the default of 60 seconds?

ole4ryb: yes, it is just to reduce close timeout, I think 10 sec is more than enough... also, wanted to test it using our perf tests...

PR #508, opened 2024-03-11
Niad-2455 test refactoring

azam145:
## What
Removing a few similar tests as they can be replaced by one test and run with different param to simulate different scenarios.

## Why

We can reduce the amount of tests (test folders) that we keep and the maintainability cost  by reducing the tests and adjusting the params before the run time to test different scenarios.

## Type of change

## Checklist:


README.md
```diff
@@ -137,3 +137,33 @@ In particular this means that this project may not depend on GPL-licensed or AGP
 as these would violate the terms of those libraries' licenses.
 
 The contents of this repository are protected by Crown Copyright (C).
+
+## Performance
+The performance of PS Adaptor was tested with JMeter tool. The use case that was used was the simulation of the patient transfer request
```
MartinWheelerMT, line 142: "the use case that was used..." reads a bit strangely, maybe change to "The use case tested was the...."

README.md
```diff
@@ -137,3 +137,33 @@ In particular this means that this project may not depend on GPL-licensed or AGP
...
+where we were sending Ehr records data to PS Adaptor which was expected to create a bundle of it. We were looking to observe how PS Adaptor
```
MartinWheelerMT, line 143: Maybe the whole thing should read more: 
"The performance of PS Adaptor was tested with JMeter tool. The use case tested was the simulation of the patient transfer request.  This was tested by sending EHR record requests to the PS Adaptor, and we expected to receive a bundle back.  We used these tests to observe how the PS Adaptor handles a heavy workload and to profile the CPU and memory usage during the testing process.

README.md
ole4ryb: thanks @MartinWheelerMT , sounds better indeed...

README.md
ole4ryb: thanks, looks better.

test-suite/non-functional-tests/test-scenario/README.md
```diff
@@ -1,31 +1,59 @@
...
-### Test details
+## Each Sucessful transfer contains
 
-120 Transfers requested - 100 expected to succeed, 20 expected to fail, each successful transfer contains:
 * 1 RCMR_IN030000UK06 message
 * 1 COPC_IN000001UK01 index message
 * 2 COPC_IN000001UK01 Fragment messages
 
-### Prerequisites 
+## Test Scenario 1
```
adrianclay, line 9: Nice clean up of the different folders.

I think it would be useful to know in a bit more detail how to specifically set up these scenarios using the JMeter file. Perhaps saying which `properties` to set with each value?

<pre>
## Test Scenario 1

12 patient transfers x 1 thread

12 Transfers requested - 10 expected to succeed, 2 expected to fail. 

```
thing = 1
blah = 100
```
</pre>

adrianclay (approved):

PR #530, opened 2024-03-22
Niad-3067: duplicated ObservationStatement entries for the same Observation

ole4ryb:
## What

ObservationStatement with "..." pattern at the end of pertinentAnnotation has been merged with one that has the full description of the same line.

## Why

There should be full description of pertinentAnnotation included instead of the truncated one.

## Type of change

## Checklist:


adrianclay:
You sure this code meets Acceptance Criteria 2 of the ticket @ole4ryb / @MartinWheelerMT ?

adrianclay:
I also wonder if over 600 lines of XML is necessary, or whether it could be cut down smaller.

ole4ryb:
> I also wonder if over 600 lines of XML is necessary, or whether it could be cut down smaller.

There are under 600 lines of XML :) anyway, I will check and see if it can be reduced.

ole4ryb:
> You sure this code meets Acceptance Criteria 2 of the ticket @ole4ryb / @MartinWheelerMT ?

Yes, I've run a test locally using Postman and see only one condition generated.

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/ConditionMapper.java
```diff
@@ -454,6 +515,29 @@ private List<RCMRMT030101UKEhrComposition> getCompositionsContainingLinkSets(RCM
...
+        RCMRMT030101UKObservationStatement referencedObservationStatement) {
+
+        List<RCMRMT030101UKObservationStatement> observationStatements = ehrExtract.getComponent().stream()
+            .map(RCMRMT030101UKComponent::getEhrFolder)
+            .map(RCMRMT030101UKEhrFolder::getComponent)
+            .flatMap(Collection::stream)
+            .map(RCMRMT030101UKComponent3::getEhrComposition)
+            .flatMap(ehrComposition -> ehrComposition.getComponent().stream())
+            .flatMap(CompoundStatementResourceExtractors::extractAllObservationStatements)
+            .toList();
```
adrianclay, line 529: Does this code satisfy this condition?

```
Be a child of the same EhrComposition
```

adrianclay: Also the condition:
```
The replaced ObservationStatement must have a single pertinentInformation / pertinentAnnotation which ends in ...
```

adrianclay: I think I can see that final condition being met on line 253, but the downside of this approach, is that the check happens after the `findFirst` call which means that we could be missing the `ObservationStatement` if the actual match isn't the first.

ole4ryb: I think we covered it with a test when the ordering is different and it works OK.

ole4ryb: > Also the condition:
> 
> ```
> The replaced ObservationStatement must have a single pertinentInformation / pertinentAnnotation which ends in ...
> ```

Yes, we've got it confirmed with tests.

PR #544, opened 2024-03-25
NIAD-2848: System identifier for DiagnosticReport should be a URN

MartinWheelerMT:
## What

Updated identifier mapping when mapping `DiagnosticReport` to ensure that that PMIP OID's are mapped as a URN (identifier code is prefixed with `"urn:oid:"`)

## Why

Identifier extension system does not contain a valid value, it should be a URN, in the form `"urn:oid:2.16.840.1.113883.2.1.4.5.5"`

## Type of change

## Checklist:


MartinWheelerMT:
To merged after changes in PR #543 are merged

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
     - or else EhrComposition effectiveTime - high
     - or else EhrComposition effectiveTime - low
 
+### Fixed
```
adrianclay, line 22: This feels like a breaking change, because NMEs using the adaptor will need to update their importing routines to look for this new "system" value.

I'm aware that we have this URN/Not a URN issue elsewhere.

If we were to make this breaking change, I'd suggest we make all the breaking changes at once to minimise pain for our users.

Thoughts?

ole4ryb (approved):

CHANGELOG.md
MartinWheelerMT: This has now been addressed, once this pull request has been approved / merged, I will need to let Emile know to ready the changes on the Medicus side for release

common/src/main/java/uk/nhs/adaptors/common/util/CodeableConceptUtils.java
azam145, line 1: Tries to convert to Urn otherwise returns original string ?
OidUtil.tryParseToUrn(system).orElse(system)

common/src/main/java/uk/nhs/adaptors/common/util/OidUtil.java
azam145, line 1: Is there no limit to length of oid string ?

azam145 (changes requested):

common/src/main/java/uk/nhs/adaptors/common/util/OidUtil.java
MartinWheelerMT: No limit, the regex for the OID check is taken from the GP2GP Specification

common/src/main/java/uk/nhs/adaptors/common/util/CodeableConceptUtils.java
MartinWheelerMT: Yes thats correct, it's following the pattern baeldung have for their TryParse... methods

azam145 (approved):

PR #616, opened 2024-06-13
Update ObservationMapper to ensure AvailabilityTime is used for Issued, when available

MartinWheelerMT:
## What

Update ObservationMapper to ensure that the availabilityTime is used when available, when mapping ObservationStatements and ObservationRequests.
Add Unit tests to ObservationMapperTest for this updated functionality.

## Why

GP Connect specifies that the issued date on ObservationStatement should use the following to determine its value:

`The audit trail timestamp representing when the data was recorded.`

It has been reported under NIAD-2911 that it has been identified that the date & time associated with Investigation Results is incorrect.

## Type of change

## Checklist:

### DIFF
--- a/gp2gp-translator/src/integrationTest/resources/json/expectedBundle.json
+++ b/gp2gp-translator/src/integrationTest/resources/json/expectedBundle.json
@@ -15,7 +15,7 @@
       },
       "identifier": [ {
         "system": "https://fhir.nhs.uk/Id/nhs-number",
-        "value": "9194544435"
+        "value": "{{nhsNumber}}"
       } ],
       "managingOrganization": {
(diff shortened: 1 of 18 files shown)


gp2gp-translator/src/integrationTest/resources/json/expectedBundle.json
```diff
@@ -2364,7 +2364,7 @@
       }, {
         "text": "Action Date: 2005-04-06"
       }, {
-        "text": "Test request statement text\n                                        New line"
+        "text": "Test request statement text                                        New line"
```
adrianclay, line 2367: Do you know why these new lines have been removed?

adrianclay (approved):

PR #623, opened 2024-06-19
NIAD-3054: Also ignore discontinuation reason where the originalText is prefix of pertinentInformation

adrianclay:
## Why

Incumbent can also include additional comments after the original reason, let's make sure we remove the duplication in that case too.

## Type of change

## Checklist:

### DIFF
--- a/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/medication/MedicationRequestPlanMapper.java
+++ b/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/medication/MedicationRequestPlanMapper.java
@@ -172,12 +172,6 @@ public class MedicationRequestPlanMapper {
     }
 
     private String extractTermText(RCMRMT030101UKDiscontinue discontinue) {
-        String originalText = "";
-        if (discontinue.hasCode() && discontinue.getCode().hasOriginalText()) {
-            originalText = discontinue.getCode().getOriginalText();
-        }
-
-
         var pertinentInfo = discontinue.getPertinentInformation()
                 .stream()
                 .map(RCMRMT030101UKPertinentInformation2::getPertinentSupplyAnnotation)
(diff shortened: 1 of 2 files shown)


gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/mapper/medication/MedicationRequestPlanMapperTest.java
```diff
@@ -551,6 +551,160 @@ public void When_MappingDiscontinue_With_MissingPertinentInformation_Expect_Defa
...
+        var statusReason = (CodeableConcept) statusReasonExt.get(0).getValue();
+
+        assertAll(
+            () -> assertThat(statusExt).hasSize(1),
+            () -> assertThat(statusReasonExt).hasSize(1),
+            () -> assertThat(statusReason.getText()).isEqualTo("Prescribing error, incorrect dosage")
+        );
+    }
+
+    @Test void When_MappingDiscontinue_With_CodingOriginalTextIsDuplicatedAsUpperCasePrefixInPertinentInformation_Expect_DisplayedTwice() {
```
ole4ryb, line 605: The tests are very similar, so they can be converted into the parameterized ones which should save a lot of code lines and make it more concise...

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/medication/MedicationRequestPlanMapper.java
```diff
@@ -187,11 +181,15 @@ private String extractTermText(RCMRMT030101UKDiscontinue discontinue) {
 
         var stringBuilder = new StringBuilder();
```
ole4ryb, line 182: Could we name the stringBuilder variable to be more in line with the method name and the expected result?

ole4ryb (approved):

PR #647, opened 2024-06-24
 [NIAD-2988] Lowercase Conversation ID for getMigrationRequest()

martin-nhs:
## What

Previously:
- **Given** a POST request made to the `/$gpc.ack` endpoint.
- **When** a request header is passed for `conversationId` with a lowercase UUIDv4.
- **Then** the response would be status 500 (Internal Server Error).

Now:
- **Given** a POST request made to the `/$gpc.ack` endpoint.
- **When** a request header is passed for `conversationId` with a lowercase OR uppercase UUIDv4.
- **Then** the response would be status 200 (OK).

## Why

This bug could cause confusion and questions from NMEs as to why the endpoint doesn’t work.

## Type of change

## Checklist:

### DIFF
--- a/gpc-api-facade/src/integrationTest/java/uk/nhs/adaptors/pss/gpc/controller/AcknowledgeRecordControllerIT.java
+++ b/gpc-api-facade/src/integrationTest/java/uk/nhs/adaptors/pss/gpc/controller/AcknowledgeRecordControllerIT.java
@@ -3,7 +3,9 @@ package uk.nhs.adaptors.pss.gpc.controller;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
 import org.junit.jupiter.params.provider.EnumSource;
+import org.junit.jupiter.params.provider.MethodSource;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc;
 import org.springframework.boot.test.context.SpringBootTest;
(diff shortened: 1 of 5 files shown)


gpc-api-facade/src/main/java/uk/nhs/adaptors/pss/gpc/service/AcknowledgeRecordService.java
```diff
@@ -26,7 +26,7 @@ public class AcknowledgeRecordService {
     public Boolean handleAcknowledgeRecord(
             @NotNull @NotEmpty String conversationId,
             @NotNull @NotEmpty String confirmationResponseString) {
-        var patientMigrationRequest = patientMigrationRequestDao.getMigrationRequest(conversationId);
+        var patientMigrationRequest = patientMigrationRequestDao.getMigrationRequest(conversationId.toLowerCase());
```
adrianclay, line 29: Hmm, strange as I'm pretty sure we're [uppercasing the conversation ID](https://github.com/NHSDigital/nia-patient-switching-standard-adaptor/blob/main/gpc-api-facade/src/main/java/uk/nhs/adaptors/pss/gpc/service/PatientTransferService.java#L39-L40) when it's inserted.

martin-nhs: You're correct, my bad!

gpc-api-facade/src/integrationTest/java/uk/nhs/adaptors/pss/gpc/controller/AcknowledgeRecordControllerIT.java
```diff
@@ -138,9 +152,20 @@ public void sendAcknowledgeRequestWithMissingConfirmationResponseHeader() throws
                 .andExpect(content().json(expectedResponseBody));
     }
 
+    // Helper Methods
```
ole4ryb, line 155: I think we don't need this comments as they a bit clutter it here, the method is simple and it is very easy to see that it is a preparing/helper method for the test.

ole4ryb (approved):

PR #684, opened 2024-07-02
NAID-3122 - Add `confidentialityCode` support to `AllergyIntoleranceMapper.java`

martin-nhs:
## What

Given the following XML which is parsed by the `AllergyIntolleranceMapper`:

```xml
<confidentialityCode
  code="NOPAT"
  codeSystem="2.16.840.1.113883.4.642.3.47"
  displayName="no disclosure to patient, family or caregivers without attending provider's authorization" />
```

When the element is mapped, then expect the bundle to contain:

```json
{
  "meta": {
    "security": [
      {
        "system": "http://hl7.org/fhir/v3/ActCode",
        "code": "NOPAT",
        "display": "no disclosure to patient, family or caregivers without attending provider's authorization"
      }
    ]
  }
}
```

## Why

Please include details of the reasoning for these changes

## Type of change

## Checklist:


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AllergyIntoleranceMapper.java
```diff
@@ -88,14 +92,18 @@ private AllergyIntolerance mapAllergyIntolerance(RCMRMT030101UKEhrComposition eh
                 .getId()
                 .getRoot();
 
+        final Meta meta = CONFIDENTIALITY_CODE_PRESENT.test(observationStatement) ?
```
adrianclay, line 95: I _suspect_ this refactoring won't be the most desirable once we've implemented a few of these mappers, but given my mindset of refactoring when the duplication is indeed visible and a problem, let's leave it as it is for now.

gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/util/ConfidentialityUtil.java
```diff
@@ -0,0 +1,42 @@
...
+import org.hl7.fhir.dstu3.model.Coding;
+import org.hl7.fhir.dstu3.model.Meta;
+import org.hl7.v3.CV;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.Optional;
+
+public final class ConfidentialityUtil {
+    private ConfidentialityUtil() { }
```
ole4ryb, line 12: I think it would be good to break a line after line 11 otherwise it is very difficult to read such a class.
Also, it would be good to have a line break between the constructor and the static variable.

ole4ryb (approved):

ole4ryb (approved):

PR #722, opened 2024-07-22
NIAD-2851: Remove unexplained call to handleObservationStatement

adrianclay:
## Why

A BATTERY CompoundStatement can only have two types of component inside it according to the GP2GP specification. These are:

- A CLUSTER CompoundStatement
- An ObservationStatement

The existing code within handleBatteryCompoundStatement deals with each of those cases already. The additional code removed in this change appears to deal with any type of CompoundStatement. Given that the only CompoundStatement which is valid to appear here is a CLUSTER, and the code for handleClusterCompoundStatement already calls handleObservationStatement, there is no point to this code.

By removing this duplicate call to handleObservationStatement we also fix the issue where Observations are being generated with two identical "category" values of "Laboratory".

Furthermore, this code is completely untested. It is a liability in its current state, so the removal of it reduces the chance of it breaking and introducing more unwelcome unintended side-effects.

## Type of change

Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/diagnosticreport/SpecimenCompoundsMapper.java
+++ b/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/diagnosticreport/SpecimenCompoundsMapper.java
@@ -230,19 +230,6 @@ public class SpecimenCompoundsMapper {
             );
         }
 
-        batteryCompoundStatement.getComponent().stream()
-            .filter(RCMRMT030101UKComponent02::hasCompoundStatement)
-            .map(RCMRMT030101UKComponent02::getCompoundStatement)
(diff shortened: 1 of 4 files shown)


gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/diagnosticreport/SpecimenCompoundsMapper.java
```diff
@@ -230,19 +230,6 @@ private void handleBatteryCompoundStatement(RCMRMT030101UKCompoundStatement spec
             );
         }
 
-        batteryCompoundStatement.getComponent().stream()
```
adrianclay, line 233: So on second look, line 224 has a filter "CLUSTER_CLASSCODE". So the duplication _only_ occurrs when the battery has clusters inside it.

ole4ryb (approved):

PR #724, opened 2024-07-25
NIAD-3138: password encoding

ole4ryb:
## What

Password percent-encoding 

## Why

Sometimes passwords can include '%' special characters which could result in "psql: error: invalid percent-encoded token...". So in order to mitigate such a situation the password percent-encoding has been introduced by this PR.

## Type of change

## Checklist:

### DIFF
--- a/snomed-database-loader/load_release-postgresql.sh
+++ b/snomed-database-loader/load_release-postgresql.sh
@@ -42,7 +42,6 @@ if [[ $1 == *uk_sct2mo* ]]; then
 fi
 
 stringencoder() {
-    # urlencode <string>
     local length="${#1}"
     for (( i = 0; i < length; i++ )); do
         local c="${1:i:1}"
--- a/snomed-database-loader/load_release-postgresql.sh
+++ b/snomed-database-loader/load_release-postgresql.sh
@@ -41,7 +41,20 @@ if [[ $1 == *uk_sct2mo* ]]; then
 	isMonolith=true
 fi
 
-databaseUri="postgresql://${PS_DB_OWNER_NAME}:${POSTGRES_PASSWORD}@${PS_DB_HOST}:${PS_DB_PORT}/${dbName}"
+stringencoder() {
+    # urlencode <string>
+    local length="${#1}"
+    for (( i = 0; i < length; i++ )); do
+        local c="${1:i:1}"
+        case $c in
+            [a-zA-Z0-9.~_-]) printf "$c" ;;
+            *) printf '%%%02X' "'$c" ;;
+        esac
+    done
+}
+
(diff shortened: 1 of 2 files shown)


snomed-database-loader/load_release-postgresql.sh
```diff
@@ -41,7 +41,20 @@ if [[ $1 == *uk_sct2mo* ]]; then
 	isMonolith=true
 fi
 
-databaseUri="postgresql://${PS_DB_OWNER_NAME}:${POSTGRES_PASSWORD}@${PS_DB_HOST}:${PS_DB_PORT}/${dbName}"
+stringencoder() {
+    # urlencode <string>
+    local length="${#1}"
+    for (( i = 0; i < length; i++ )); do
+        local c="${1:i:1}"
```
martin-nhs, line 48: What is `c` in this instance?

snomed-database-loader/load_release-postgresql.sh
```diff
@@ -41,7 +41,20 @@ if [[ $1 == *uk_sct2mo* ]]; then
 	isMonolith=true
 fi
 
-databaseUri="postgresql://${PS_DB_OWNER_NAME}:${POSTGRES_PASSWORD}@${PS_DB_HOST}:${PS_DB_PORT}/${dbName}"
+stringencoder() {
+    # urlencode <string>
```
martin-nhs, line 45: Is this comment necessary?

ole4ryb: not really, I can delete it

snomed-database-loader/load_release-postgresql.sh
martin-nhs: Awesome! Could we possibly change `c` to `character` assuming this isn't a reserved word in bash - think it could be a helpful readability wise :D

snomed-database-loader/load_release-postgresql.sh
```diff
@@ -41,7 +41,19 @@ if [[ $1 == *uk_sct2mo* ]]; then
 	isMonolith=true
 fi
 
-databaseUri="postgresql://${PS_DB_OWNER_NAME}:${POSTGRES_PASSWORD}@${PS_DB_HOST}:${PS_DB_PORT}/${dbName}"
+stringencoder() {
```
martin-nhs, line 44: Last thing from me would be - `stringecoder` is a little confusing from an outside-in perspective, some questions I initially had:

* Encoded into what form? (Percent encoding)[https://developer.mozilla.org/en-US/docs/Glossary/Percent-encoding]?

martin-nhs (approved):

PR #727, opened 2024-07-26
Upgrade to latest Postgres

adrianclay:
## Why

According to _sources_, Postgres 16 is a lot faster for importing data via COPY.

https://pganalyze.com/blog/5mins-postgres-16-faster-copy-bulk-load

Given that our pipeline is slowed down by the DB SNOMED loader, this should make our builds faster.

## Type of change

Internal change (non-breaking change with no effect on the functionality affecting end users)

## Checklist:

### DIFF
--- a/docker/db/Dockerfile
+++ b/docker/db/Dockerfile
@@ -1,1 +1,1 @@
-FROM postgres:14.0
+FROM postgres:16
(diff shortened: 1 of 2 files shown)


adrianclay:
I couldn't see any change in performance comparing this version to existing version in Jenkins so closing...

PR #752, opened 2024-08-08
NIAD-3148 - Immunizations outside of consultation are mapped to Observation

martin-nhs:
## What

This PR resolves a bug where an immunization was incorrectly mapped to an Observation due to a SNOMED code mismatch. During debugging, it was identified that the code being processed was a `Description ID` rather than a `Concept ID`, which caused the mapping error. To address this, the immunization codes view has been modified to include a join on the `Description ID`. This update allows for accurate querying and validation, ensuring the correct mapping of immunization data.

## Why

More details can be found on the [ticket](https://nhse-dsic.atlassian.net/browse/NIAD-3148).

## Type of change

## Checklist:


gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/TestUtility.java
```diff
@@ -28,6 +32,32 @@ public static CV createCv(String code) {
...
+
+    /**
+     * An EHR Extract has a cardinality of one to many components, each component (based
+     * of the UK05 schema) can contain one and only one EHR Folder. This utility method provides
+     * a means of extracting ALL components from within a target EHR Folder.
+     * @param extract The EHR Extract.
+     * @param extractComponentIndex The index of the RCMRMT030101UKComponent which houses the EHR Folder.
+     * @return A list of RCMRMT030101UKComponent3.
+     */
+    public static List<RCMRMT030101UKComponent3> getEhrFolderComponent(RCMRMT030101UKEhrExtract extract,
```
martin-nhs, line 51: Let's pluralise this!

```suggestion
    public static List<RCMRMT030101UKComponent3> getEhrFolderComponents(RCMRMT030101UKEhrExtract extract,
```

gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/TestUtility.java
```diff
@@ -28,6 +32,32 @@ public static CV createCv(String code) {
         return createCv(code, "", "");
     }
 
+    public static CD createCd(String code, String codeSystem, String displayName) {
+        final CD cd = new CD();
```
martin-nhs, line 36: Would it be worth implementing the [builder pattern](https://refactoring.guru/design-patterns/builder) for `CD` and `CV` types, seems as if it could be handy.

gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/TestUtility.java
```diff
@@ -1,19 +1,23 @@
...
     private TestUtility() { }
 
     public static final Function<RCMRMT030101UKEhrExtract, RCMRMT030101UKEhrComposition> GET_EHR_COMPOSITION =
         extract -> extract
-            .getComponent().get(0)
+            .getComponent().getFirst()
             .getEhrFolder()
-            .getComponent().get(0)
+            .getComponent().getFirst()
             .getEhrComposition();
```
martin-nhs, line 21: As I've learned more since joining NIA, this function is not great. This assumes that there is one `EHR Folder`, but no, this is not the case. According to the Message Implementation Manual, and `UK05` XSD schema - one `EHR Extract` can have many components, which can contain one and only one `EHR Folder`.

So this can definitely be improved.

Technically out-of-scope for this ticket though, to raise in a separate PR.

gp2gp-translator/src/integrationTest/java/uk/nhs/adaptors/pss/translator/DatabaseImmunizationCheckerIT.java
```diff
@@ -0,0 +1,56 @@
...
+    void When_Immunization_With_SnomedDescriptionId_Expect_True() {
+        final String immunizationDescriptionSnomedId = "67308009";
+        final RCMRMT030101UKObservationStatement observationStatement =
+            new RCMRMT030101UKObservationStatement();
+
+        observationStatement.setCode(createCd(
+            immunizationDescriptionSnomedId,
+            SNOMED_CODE_SYSTEM,
+            DISPLAY_NAME
+        ));
```
martin-nhs, line 33: Not a fan of calling methods in methods, let's take this out into its own variable, likewise for the rest of em'.

gp2gp-translator/src/integrationTest/java/uk/nhs/adaptors/pss/translator/DatabaseImmunizationCheckerIT.java
```diff
@@ -0,0 +1,56 @@
...
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+import uk.nhs.adaptors.pss.translator.util.DatabaseImmunizationChecker;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static uk.nhs.adaptors.pss.translator.TestUtility.createCd;
+
+@ExtendWith(SpringExtension.class)
+@SpringBootTest
+class DatabaseImmunizationCheckerIT {
+    private static final String SNOMED_CODE_SYSTEM = "2.16.840.1.113883.2.1.3.2.4.15";
```
martin-nhs, line 17: Let's be specific that this is an [OID](https://www.hl7.org.uk/standards/object-identifiers-oids/hl7-uk-issued-oids/#:~:text=2.16.840.1.113883.2.1.3.2.4.15).

```suggestion
    private static final String SNOMED_CODE_SYSTEM_OID = "2.16.840.1.113883.2.1.3.2.4.15";
```

db-connector/src/main/java/uk/nhs/adaptors/connector/dao/ImmunizationSnomedCTDao.java
```diff
@@ -9,7 +9,7 @@
 public interface ImmunizationSnomedCTDao {
     @SqlQuery("select_immunization_concept_id")
     @UseClasspathSqlLocator
-    ImmunizationSnomedCT getImmunizationSnomednUsingConceptId(@Bind("conceptId") String conceptId);
+    ImmunizationSnomedCT getImmunizationSnomedUsingConceptOrDescriptionId(@Bind("sctid") String sctId);
```
martin-nhs, line 12: Hmm, `sctid` feels a bit lazy. As a developer joining the project and taking a look at that it would mean nothing to them. Let's rename this to `snomedId`?

martin-nhs, line 12: Worth noting too that the return type of `ImmunizationSnomedCT` is completely useless, this method is only called within `DatabaseImmunizationChecker`, in turn only returning a `boolean`. Why not just have this return a `boolean`?

Again, technically **out of scope** of this ticket, though will be tidied in a separate PR.

gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/TestUtility.java
martin-nhs: **TODO** another PR.

gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/TestUtility.java
martin-nhs: **TODO** another PR.

db-connector/src/main/java/uk/nhs/adaptors/connector/dao/ImmunizationSnomedCTDao.java
martin-nhs: **TODO** another PR.

snomed-database-loader/create-database-postgres.sql
```diff
@@ -70,8 +70,9 @@ WITH RECURSIVE immunization_heirarchy AS (
...
     WHERE r.typeId = '116680003' -- relationshipType (typeId) is 'IsA' (child of)
 )
-SELECT conceptId
-FROM immunization_heirarchy;
+SELECT ih.conceptId, ds.id as descriptionid
+FROM immunization_heirarchy ih
+JOIN snomedct.description_s ds on ih.conceptId = ds.conceptid;
 
 CREATE INDEX immunization_codes_conceptid_idx ON immunization_codes
     USING btree (conceptid);
```
adrianclay, line 78: Sorry for the randomly timed comment, but it popped into my head. This `btree` won't be as effective as it was previously now that the SELECT query we perform needs to perform a search across two columns.

If we were able to combine the `conceptId` and `descriptionId` into one column, and have a single btree for both, I suspect that would be the most performant solution.

martin-nhs: I agree, great spot! Will address this ASAP.

adrianclay (approved):

PR #870, opened 2024-10-09
Gradle and Spring upgrade in Mock Spine Mhs Outbound

ole4ryb:
## What

Gradle and Spring boot version upgrade

## Why

The Mock Spine MHS Outbound module in the test suite was using outdated Gradle and Spring Boot versions, which caused it to lag behind other modules. This PR updates the module to align with the versions used by other components, ensuring compatibility across Java versions and preventing any potential issues related to version mismatches.

## Type of change

## Checklist:


test-suite/mock-spine-mhs-outbound/build.gradle
```diff
@@ -1,7 +1,7 @@
 plugins {
-    id 'org.springframework.boot' version '2.7.1'
-    id 'io.spring.dependency-management' version '1.0.11.RELEASE'
-    id 'io.freefair.lombok' version '6.2.0'
+    id 'org.springframework.boot' version '3.3.3'
```
martin-nhs, line 2: Wouldn't it be better if we allow the patch level to be the latest, leaving the major and minor levels as-is? E.g. `3.3.x`?

ole4ryb: I just took it from other project services where we are using minor version which currently is ver. 3.3.3,

test-suite/mock-spine-mhs-outbound/build.gradle
```diff
@@ -1,7 +1,7 @@
...
+    id 'io.spring.dependency-management' version '1.1.6'
```
martin-nhs, line 3: Same as the comment above, `1.1` should suffice - if we choose that this is OK to do so.

ole4ryb: again, I am aligning it with the other modules, like gp2gp-translator or common or db-connector... they are all using ver. 1.1.6. I think the general rule is that it is better to use more specific version given we are already have other modules using the same version.

test-suite/mock-spine-mhs-outbound/build.gradle
martin-nhs: OK, no biggie - just thought I would highlight it 😄

test-suite/mock-spine-mhs-outbound/build.gradle
martin-nhs: Okay doak! Makes sense, just thought I'd highlight!

martin-nhs (approved):

PR #927, opened 2024-11-15
Fix bug where if the name element was omitted for an `agentPerson` then an exception would be thrown

MartinWheelerMT:
## What

Fixed a bug where if the name element was omitted for an `agentPerson` (which is a valid use case) then an exception would be thrown.  Instead, this now produces a `Practitioner` with a family name of `"unknown"` as documented.

Added Unit Test for the above functionality.

## Why

When an `Agent` is provided in the XML with a `agentPerson` element but without a nested `name` element: 

```xml
<Agent classCode="AGNT">
    <id root="6C8AC7E0-4D90-10C5-5FD3-F2C05C658E18" />
    <code code="125676002" codeSystem="2.16.840.1.113883.2.1.3.2.4.15" displayName="Person">
        <originalText>Other</originalText>
    </code>
    <agentPerson classCode="PSN" determinerCode="INSTANCE" />
</Agent>
```

Then the adaptor currently throws a BundleMappingException.  It is expected instead that this is handled gracefully, with produced `Practitioner.familyName` being set to `"unknown"`

## Type of change

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## [Unreleased]
 
+### Fixed
+
+* When provided an `agentPerson` without a `name` element, the adaptor will no longer throw an exception, but will
+  instead map the `Practitioner` with a `familyName` of `"unknown"`.
+
 ## [3.0.6] - 2024-11-11
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


adrianclay:
Code looks fine, could do with a CHANGELOG entry though.

MartinWheelerMT:
> Code looks fine, could do with a CHANGELOG entry though.

Done!

adrianclay (approved):

PR #936, opened 2024-11-19
NIAD-3220: Add Resource filter to identify "ReferralRequest to ExternalDocument LinkSets"

MartinWheelerMT:
## What

* Add Test to check verify that the provided `LinkSet` is a "Referral Request to External Document LinkSet."
* Add test XML file containing a "Referral Request to External Document LinkSet."
* Add method `isReferralRequestToExternalDocumentLinkSet` to `ResourceFilterUtil` with parameters for the `EhrExtract` and the `LinkSet` itself to identify `LinkSets` matching the criteria for a "Referral Request to External Document LinkSet."

## Why

As part of the implementation for NIAD-3220, it is necessary to identify `LinkSets` which are a "Referral Request to External Document LinkSet" in order to be able to then use this for filtering in the `ConditionMapper` to exclude the `Condition` from being mapped, and in `ReferralRequest` mapper to identify any of these LinkSets when populating the `supportingInfo` field.

A "ReferralRequest to ExternalDocument LinkSet" is defined as a `LinkSet` which matches the following criteria:

- It has a `code` with a `value` of `394776006` and a `system` of `2.16.840.1.113883.2.1.3.2.4.15` and no `qualifier` and no `originalText`
- The `conditionNamed / namedStatementRef / id[root]` is a `RequestStatement`
- It has at least 1 `component`
- All the `component / statementRef / id[root]` are references to GP2GP document/attachments.

## Type of change

## Checklist:

### DIFF
--- a/gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/util/ResourceFilterUtilTest.java
+++ b/gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/util/ResourceFilterUtilTest.java
@@ -8,6 +8,7 @@ import static uk.nhs.adaptors.pss.translator.util.XmlUnmarshallUtil.unmarshallFi
 import java.util.Objects;
 import java.util.stream.Stream;
 
+import org.hl7.v3.CR;
 import org.hl7.v3.RCMRMT030101UKComponent;
 import org.hl7.v3.RCMRMT030101UKComponent4;
 import org.hl7.v3.RCMRMT030101UKCompoundStatement;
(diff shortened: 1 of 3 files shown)


gp2gp-translator/src/test/java/uk/nhs/adaptors/pss/translator/util/ResourceFilterUtilTest.java
```diff
@@ -157,6 +162,26 @@ private static Stream<Arguments> nonSpecimenTestFiles() {
         );
     }
 
+    @Test
+    public void testIsReferralRequestToExternalDocumentLinkSet() {
```
adrianclay, line 166: Been a bit distracted, but my initial thought @MartinWheelerMT is that the following implementation
```java
public static bool isReferralRequestToExternalDocumentLinkSet() {
  return true;
}
```

Would pass the tests, so I'm not sure why your prod code is so complicated.

adrianclay (approved):

PR #952, opened 2024-12-04
NIAD-3174-migrate-to-github-actions

MartinWheelerMT:
## What

- [x] New GitHub Actions workflow which runs the same tests/checks that the current Jenkins job does.
- [x] Workflow runs on every PR, and push to main branch
- [x] The old Jenkins build is disabled to prevent the two builds conflicting, but can be re-enabled if required.
- [ ] --A timeout is set for the whole workflow at 1hour-- (not possible for github hosted runners)

## Why

To facilitate the move away from Jenkins

## Type of change

## Checklist:


adrianclay (approved):
- Don't think the facade needs SNOMED
- It might be simpler to have `gradle` handle the parallelisation instead of GitHub Actions.

PR #965, opened 2024-12-20
Run checkstyle and spotbugs against IntegrationTest in github actions

MartinWheelerMT:
## What

Added checkstylentegrationTests and spotbugsIntegrationTests to github `Test` action

## Why

Checkstyle and SpotBugs were not running against the integration tests.

## Type of change

## Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -32,6 +32,7 @@ jobs:
         run: |
           ./gradlew checkStyleMain --parallel
           ./gradlew checkstyleTest --parallel
+          /gradlew checkstyleIntegrationTest --parallel
         working-directory: ${{ inputs.path }}
 
       - name: Collect Artifacts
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -73,6 +74,7 @@ jobs:
         run: |
           ./gradlew spotbugsMain --parallel
           ./gradlew spotbugsTest --parallel
+          ./gradlew spotbugsIntegrationTest --parallel
```
adrianclay, line 77: I think you lose the benefit of parallel when executing each command sequentially like this right?!

MartinWheelerMT: I thought the --parallel was used to run any task within the specified as parallel (ie all tasks in spotbugsMain run in parallel, not that each job is run in a parallel, but I've done some benchmarking (very losely) and there doesn't appear to be any difference with it removed.

I've tried different ways of doing it (including: `./gradlew spotbugsMain spotbugsTest spotbugsIntegrationTest`) and none of them seem to make any difference at all (I ensured the gradle cache was cleared first).

I'll merge these into one line and remove the parallel as it does nothing for our jobs

adrianclay (approved):

PR #1262, opened 2026-05-06
NIAD-3453: Refactor StorageService to correctly use DI/IOC

MartinWheelerMT:
## What

* Delete StorageServiceFactory.java and StorageServiceFactoryConfig.java as unnecessary and confusing steps in the DI process.
* Introduce StorageServiceConfig to conditionally create the required service.
* Refactor AzureStorageService to now correctly inject the StorageServiceClient from the IOC container.
* Add Integration Test sfor AzureStorageService.
* Add `testcontainers` to `build.gradle` to enable use of `Azurite`.

## Why

DI was incorrectly set up and abstracted multiple times, causing the storage service to be manually created rather than correctly injected.  This was refactored so that the `bean` is now correctly using IOC.
Since its creation the AzureStorageService class has not had any test coverage, and it's in our best interest to add this.

## Type of change

## Checklist:


chiaramapellimt (approved):
LGTM

PR #1292, opened 2026-06-23
NIAD-3471 Support multiple given names in practitioner XML

MartinWheelerMT:
## What

- Changed EN.given field from String to List<String> to support multiple given name elements
- Updated AgentDirectoryMapper to iterate over all given names and add each to FHIR HumanName
- Added helper method getGivenNamesAsString() to concatenate multiple given names for text field
- Updated hasNoName() to correctly check for given names in List structure
- Updated JAXB schema test to work with List<String> given names
- Added convenience method getFirstGiven() for backward compatibility
- Added unit tests for multiple given names scenarios (with and without family name).

## Why

The adapter was unable to handle XML containing multiple <given> fields for practitioner names. When an XML part contained multiple given names like:

```XML
<name use="L">
    <given>Minire</given>
    <given>E</given>
    <family>Clarkson</family>
</name>
```

The JSON output would only include a single given name:

```JSON
{
    "name": [{
        "use": "official",
        "family": "Clarkson",
        "given": ["E"]  // Only one given name captured
    }]
}
````

**Root Cause**

The JAXB-generated schema classes in org.hl7.v3.EN (which PN extends) only supported a single String field for given names:

`protected String given`;

This meant when the XML was unmarshalled, JAXB could only capture one <given> element, even though the XML schema (datatypes.xsd) defines the name parts as a <choice> element with maxOccurs="unbounded", allowing multiple instances.

**Solution**

_**1. Modified EN Class (schema/src/main/java/org/hl7/v3/EN.java)**_

Changed the given field to support multiple values:

// Before:
`protected String given;`

// After:
`protected List<String> given;`

Updated the getter method to return a List:

```Java
public List<String> getGiven() {
    if (given == null) {
        given = new ArrayList<>();
    }
    return this.given;
}
```

Added a convenience method for backward compatibility:

```Java
public String getFirstGiven() {
    if (given != null && !given.isEmpty()) {
        return given.get(0);
    }
    return null;
}
```

**_2. Updated AgentDirectoryMapper_** (gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AgentDirectoryMapper.java)

Modified the mapper to iterate over all given names and add them to the FHIR HumanName:

```Java
private void setHumanNameValuesFromName(PN name, HumanName humanName) {
    humanName.setFamily(name.getFamily());

    // Handle multiple given names
    var givenList = name.getGiven();
    if (givenList != null && !givenList.isEmpty()) {
        for (String givenName : givenList) {
            var given = getPractitionerGiven(givenName);
            if (given != null) {
                humanName.getGiven().add(given);
            }
        }
    }

    var prefix = getPractitionerPrefix(name.getPrefix());
    if (prefix != null) {
        humanName.getPrefix().add(prefix);
    }
}
```

Added a helper method to concatenate given names for the text field when there's no family name:

```Java
private static String getGivenNamesAsString(PN name) {
    var givenList = name.getGiven();
    if (givenList != null && !givenList.isEmpty()) {
        return StringUtils.join(givenList, " ");
    }
    return null;
}
```

**_3. Updated Tests_**

Updated JaxbTest (schema/src/test/java/JaxbTest.java)

Changed the test to check for a list of given names:

// Before:
`assertThat(person.getName().getGiven()).isEqualTo("Peter");`

// After:
`assertThat(person.getName().getGiven()).hasSize(1);`
`assertThat(person.getName().getGiven().getFirst()).isEqualTo("Peter");`

**_Updated and Added Tests in AgentDirectoryMapperTest_**

Updated the test When_MapAgentDirectoryOnlyAgentPersonWithNameElementWithOnlyGiven_Expect_TextSetToGiven() to correctly test the scenario with both multiple given names and a family name.

Added new test When_MapAgentDirectoryOnlyAgentPersonWithMultipleGivenNamesNoFamily_Expect_TextSetToGiven() to test the case where there are multiple given names but no family name.

Added new test When_MapAgentWithMultipleGivenAndFamilyName_Expect_AllGivenNamesInArray() to test the exact user scenario with ID and all assertions.

**Expected Output After Fix**

For the XML example provided:

```XML
<part typeCode="PART">
    <Agent classCode="AGNT">
        <id root="CD8E40B3-6A3C-11F1-AE7C-00155D75C807"/>
        <agentPerson classCode="PSN" determinerCode="INSTANCE">
            <name use="L">
                <given>Minire</given>
                <given>E</given>
                <family>Clarkson</family>
            </name>
        </agentPerson>
    </Agent>
</part>
```

The JSON output is now:

```JSON
{
    "resource": {
        "resourceType": "Practitioner",
        "id": "CD8E40B3-6A3C-11F1-AE7C-00155D75C807",
        "meta": {
            "profile": ["https://fhir.nhs.uk/STU3/StructureDefinition/CareConnect-GPC-Practitioner-1"]
        },
        "name": [{
            "use": "official",
            "family": "Clarkson",
            "given": ["Minire", "E"]
        }]
    }
}
```

The changes are backward compatible:

* Single given names still work as before
* The convenience method getFirstGiven() provides access to the first given name
* All existing tests pass without modification (except for the assertion adjustment to work with Lists)

## Type of change

## Checklist:

Co-authored-by: GitHub Copilot <noreply@github.com>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,9 +12,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ### Added
 * Added support for mapping different EhrSupplyType (e.g. NHS prescription, OTC sale) into the Medication Statement Prescribing Agency extension
-* Added fallback for Condition.asserter to use the EHRComposition / author / agent field when EHRComposition / participant2 is absent.
-* 
+* Added fallback for Condition.asserter to use the EHRComposition / author / agent field when EHRComposition / participant2 is absent.
+
 ### Fixed
+* Fixed handling of multiple `<given>` name fields in practitioner XML to JSON mapping - now correctly captures all given names as an array in FHIR HumanName.
 * Improved error handling in SkeletonProcessingService to throw a meaningful IllegalArgumentException
   when a payload node cannot be matched to a skeleton document ID, replacing an uninformative NullPointerException.
 
--- a/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AgentDirectoryMapper.java
+++ b/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AgentDirectoryMapper.java
@@ -156,11 +156,15 @@ public class AgentDirectoryMapper {
     }
 
     private static String getGivenNamesAsString(PN name) {
-        var givenList = name.getGiven();
-        if (givenList != null && !givenList.isEmpty()) {
-            return StringUtils.join(givenList, " ");
+        if (name == null) {
+            return null;
         }
-        return null;
+
+        var givenList = name.getGiven().stream()
+            .filter(StringUtils::isNotBlank)
+            .toList();
+
+        return givenList.isEmpty() ? null : StringUtils.join(givenList, " ");
     }
 
     private static boolean hasNoName(PN name) {
--- a/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AgentDirectoryMapper.java
+++ b/gp2gp-translator/src/main/java/uk/nhs/adaptors/pss/translator/mapper/AgentDirectoryMapper.java
@@ -130,17 +130,23 @@ public class AgentDirectoryMapper {
     }
 
     private static void setTextFromAvailableNameFields(PN name, HumanName humanName) {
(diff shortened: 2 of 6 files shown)


MartinWheelerMT:
Mutations are false positives, I have tried to recreate but the mutation scenarios are not possible to recreate.

chiaramapellimt (approved):

